Issue #273, opened 2015-01-29
glmer offset problems


bbolker:
Two similar problems reported by Steve Bellan and Xavier Paoletti.

Probably some sort of internal overflow (verbose=100 shows that we are getting `nan` values), but it's hard to see immediately where they're coming from since I thought we clamped all the relevant derivative and inverse-link functions to avoid this sort of problem ...

```
library("lme4")
load(system.file("testdata","sbTobb.Rdata",package="lme4"))
glmer(infected ~ vacc + (1|cluster) + offset(log(perstime)), 
      family=binomial(link='cloglog'),  data = temp,
      verbose=100)

dd <- read.table(system.file("testdata",
                             "dataEx_Glmer.txt",package="lme4"),
                 header=TRUE)
slope <- 1.5
glmer(DLT ~ offset(slope*dose) + (1 | patid),
      family=binomial,data=dd)
```

stevencarlislewalker:
As usual I could be off here, but it looks like `clogloginv` is only bounded below one, not above zero.
```{c}
    template<typename T>
    struct clogloginv : public std::unary_function<T, T> {
	const T operator() (const T& x) const {
	    return T(std::min(1.-std::numeric_limits<T>::epsilon(),
                              pgumbel2(double(x), 0., 1., 1)));
	}
    };
```
Contrast this with `R`'s `clogloginv`,
```{r}
pmax(pmin(-expm1(-exp(eta)), 1 - .Machine$double.eps), .Machine$double.eps)
```

bbolker:
I added the lower bound and got somewhat farther:

```
*** pwrssUpdate step 0
pdev=8.27503e+09; delu_min: -28564.7; delu_max: 28308.6; delb_min: -54659; delb_max: -28089.5
*** pwrssUpdate step 1
pdev=916549; delu_min: -439.197; delu_max: 47.3528; delb_min: -1.09888e+14; delb_max: 1.73208e+14
*** pwrssUpdate step 2
pdev=716149; delu_min: -15.7878; delu_max: 35.8027; delb_min: -4.38236e+15; delb_max: 5.40432e+12
*** pwrssUpdate step 3
pdev=32184.8; delu_min: -15.7878; delu_max: 35.8027; delb_min: -4.3829e+15; delb_max: 6.92645e+13
*** pwrssUpdate step 4
pdev=32184.8; delu_min: -15.7878; delu_max: 35.8027; delb_min: -4.26761e+15; delb_max: 1.72926e+13
[...]

PR #274, opened 2015-01-31
control initial pass with nAGQ = 0

stevencarlislewalker:
A simple change to address #273.  To avoid argument creep, I put the change in `glmerControl`.  I used a pull request because I did this quickly and wanted to make sure other people would look at it.

### DIFF
--- a/R/lmer.R
+++ b/R/lmer.R
@@ -116,16 +116,18 @@ glmer <- function(formula, data=NULL, family = gaussian,
     }
 
     ## FIX ME: allow calc.derivs, use.last.params etc. if nAGQ=0
-    opt <- optimizeGlmer(devfun,
-                   optimizer = control$optimizer[[1]],
-                   ## DON'T try fancy edge tricks unless nAGQ=0 explicitly set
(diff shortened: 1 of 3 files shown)


stevencarlislewalker:
TODO: documentation

mmaechler:
Looks good to me...  but then I've never been an expert about the Hermite/Laplace/... approximation issues.  Maybe Rune.   OTOH, it's just an option and I like that you added it only to the control* argument list.

bbolker:
fine with me ...

runehaubo:
Looks good to me as well. If the `nAGQ=0` option has a tendency to underestimate the RE variance (this is my expectation based on similarity with h-likelihood methods, but I haven't got concrete evidence), it may generally in cases like this provide poor starting-values for these parameters, so a general way to by-pass the `nAGQ=0` initial step seems like a good idea.

stevencarlislewalker:
Thanks @runehaubo.  You may add one more data point to your conjecture, as `nAGQ = 0` tends to underestimate the variance in this problem as well (issue #273).  It's helpful to hear that you have had similar experiences.

PR #286, opened 2015-03-16
Consistently use \Lambda for the left factor of relative covariance

jpritikin:
Does this look okay?


stevencarlislewalker:
Yes this is great, thank you!  Sorry that it took so long.

Issue #331, opened 2015-08-24
More warnings that can't be caught


pitakakariki:
NB: I don't have enough time today to make this report reproducible, but I can come back and do that later this week if you think that would be helpful.

This is really a bug in R, rather than lme4. It's possible to put a workaround into lme4, but it might be more sensible to escalate this to R itself.

I only ran into this because I was using a "year" variable that wasn't centred as a fixed effect, so it's unlikely to cause anyone too many headaches.

---

This is another case of https://github.com/lme4/lme4/issues/304. See also [this Stack Overflow post](http://stackoverflow.com/questions/20268021/how-to-catch-warnings-sent-during-s4-method-selection).

In the definition for `summary.merMod`:

```
    coefs <- cbind("Estimate" = coefs,
               "Std. Error" = sqrt(diag(vcov(object,
               use.hessian = use.hessian))))
```

Because `diag` is an S4 method, any warnings from vcov won't be caught by calling handlers.

pitakakariki:
I ran into this bug again today. The underlying bug in R is at least two years old! Anyone here have enough influence to get its priority increased?

bbolker:
can you provide a link to a corresponding R bug report?

pitakakariki:
I think it's this one: https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16111

pitakakariki:
Thanks @mmaechler for chasing this up. Do you want to close this as an issue here, or leave it until the underlying R bug is fixed?

mmaechler:
> Do you want to close this as an issue here, or leave it until the
    > underlying R bug is fixed?

I'd keep the issue open.  After having browsed and partly read
the C code in R,  I'm not 100% sure that our issue is exactly
corresponding to the R bug mentioned, or if it is just related.

Also it is not quite clear which future versions of R will have
this fixed  (R >= 3.3.0  for sure).

pitakakariki:
The underlying R bug has been fixed.

PR #341, opened 2015-12-13
Workaround for an S4 warnings bug.

pitakakariki:
A workaround to fix https://github.com/lme4/lme4/issues/331

### DIFF
--- a/R/lmer.R
+++ b/R/lmer.R
@@ -2276,9 +2276,9 @@ summary.merMod <- function(object,
     famL <- famlink(resp = resp)
     p <- length(coefs <- fixef(object))
 
+    stdError <- sqrt(diag(vcov(object, use.hessian = use.hessian)))
     coefs <- cbind("Estimate" = coefs,
-                   "Std. Error" = sqrt(diag(vcov(object,
-                   use.hessian = use.hessian))))
+                   "Std. Error" = stdError)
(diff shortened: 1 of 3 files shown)


bbolker:
Thanks! Seems to have broken a travis check though.  I will have to see what's going on ...

pitakakariki:
The errors have to do with `glmer.nb`. I don't think those tests are passing in the main branch either at the moment?

bbolker:
I see your point, but as of the last Travis run it *was* working ...  I'll
try restarting the build and see if it fails this time ...

pitakakariki:
Still passing then, that's pretty strange.

Slightly different environments for the two travis-ci checks, e.g. original has R 3.2.2, pull request has 3.2.3. I don't think it's the R version though, I've had `testthat::test_package("lme4")` fail with both 3.2.2 and in dev R on my laptop. (That's for "lme4/lme4", not "pitakakariki/lme4").

bbolker:
I do know that the problems with `glmer.nb` are *very* sensitive.  It doesn't surprise me that the tests fail on another platform.  Hopefully I will find a day or two to dig into this in the very near future ...

pitakakariki:
Looks like I've messed up this pull request by pushing additional commits to my fork. I'll close it and resubmit.

Issue #319, opened 2015-07-24
glmer.nb()  still doubtful 


mmaechler:
After all the closed issues about `glmer.nb()` problems which I have fixed now, notably the #287 one where we now do get correct estimates of fixed and random effects,...

there's still a basic misfunctioning,  notably exposed by the simple artificial example in `example(glmer.nb)`  -  get the latest git version of today!

The simulated example actually has "zero" random effects (not in the model, but the simulated data),
and hence `MASS :: glm.nb(.)`  works quite fine, finding "good" fixed effect estimates (and neg.bin dispersion parameter `theta`).
For some reason, in that example, the fixed effects estimated from our  `glmer.nb()` are ridiculously wrong.

bbolker:
yikes.  at least we're digging into it.  thanks for checking.

At some point it will/would be interesting to do a systematic comparison of lognormal-Poisson vs NB, along the lines of the reference below (warning: I haven't looked at it closely, I don't actually know what I think of it!)

@article{harrison_comparison_2015,
	title = {A comparison of observation-level random effect and {Beta}-{Binomial} models for modelling overdispersion in {Binomial} data in ecology \&amp; evolution},
	volume = {3},
	issn = {2167-8359},
	url = {https://peerj.com/articles/1114},
	doi = {10.7717/peerj.1114},
	language = {en},
	urldate = {2015-07-21},
	journal = {PeerJ},
	author = {Harrison, Xavier A.},
	month = jul,
	year = {2015},
	pages = {e1114}
}

bbolker:
I can't reproduce.  With most recent git version and running the code below I get `[1] "Mean relative difference: 4.158119e-06"` for the comparison of fixed effects.  (on both 32-bit Ubuntu 12.04 r-devel and 64-bit MacOS R 3.1.2)

```
set.seed(101)
dd <- expand.grid(f1 = factor(1:3),
                  f2 = LETTERS[1:2], g=1:9, rep=1:15,
                  KEEP.OUT.ATTRS=FALSE)
summary(mu <- 5*(-4 + with(dd, as.integer(f1) + 4*as.numeric(f2))))
dd$y <- rnbinom(nrow(dd), mu = mu, size = 0.5)

require("MASS")
m.glm <- glm.nb(y ~ f1*f2, data=dd, trace=TRUE)
summary(m.glm)
library(lme4)
m.nb <- glmer.nb(y ~ f1*f2 + (1|g), data=dd, verbose=TRUE)
[...]

PR #343, opened 2015-12-15
Reset offsets after stage 2 boundary check

pitakakariki:
I think this might fix https://github.com/lme4/lme4/issues/319.

I'm not 100% sure what's going on with `offset`, `baseOffset`, and needing to replace one with the other after you call `devfun`. But it looks like not doing this after `optimizeGlmer` calls `check.boundary` is what's been causing the numerical errors in `glmer.nb`.

Anyway, it's passing Travis on my fork, but someone should double check to see if this fix actually makes sense.

### DIFF
--- a/R/modular.R
+++ b/R/modular.R
@@ -771,10 +771,13 @@ optimizeGlmer <- function(devfun,
     }
     if (restart_edge) ## FIXME: implement this ...
         stop("restart_edge not implemented for optimizeGlmer yet")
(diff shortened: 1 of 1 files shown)


bbolker:
this is very exciting!

pitakakariki:
I find that `return(.)` speeds up my reading comprehension, so the trade off is worthwhile in my own code. I'll keep that in mind for other projects though.

PR #344, opened 2015-12-15
Workaround for an S4 warnings bug.

pitakakariki:
Same one as before, but without the extraneous commits.


pitakakariki:
Still fails Travis, but that's the same `glmer.nb` stuff again.

PR #364, opened 2016-02-23
Fix typo in man/modular.Rd

davidkretch:
Hi, I have made a very small change to fix what I think is a typo in the 'Modular Functions for Mixed Model Fits' documentation:

Original: 'both the fixed-effects and fixed-effects part'
Fix: 'both the fixed-effects and random-effects parts'


bbolker:
thanks!

Issue #400, opened 2016-11-03
Potentially confusing behaviour for simulate.formula with binomial models.


pitakakariki:
This is some of the example code from `?simulate.merMod`.

```
gm1 <- glmer(cbind(incidence, size - incidence) ~ period + (1 | herd),
             data = cbpp, family = binomial)

params <- list(theta=0.5,beta=c(2,-1,-2,-3))
simdat <- with(cbpp,expand.grid(herd=levels(herd),period=factor(1:4)))
simdat$size <- 15
simdat$incidence <- sample(0:1,size=nrow(simdat),replace=TRUE)

form <- formula(gm1)[-2]

simulate(form,newdata=simdat,family=binomial,
    newparams=params)[[1]]
# theta parameter vector not named: assuming same order as internal vector
# beta parameter vector not named: assuming same order as internal vector
#  [1] 1 1 1 0 1 1 0 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 0 1 1 1 0 0 1 1 0 0 1 0 1 0 0 0 1 1 0 0 1 1 0 0 1 0 0 1 0 1
[56] 0 0 0 1 0
```

It's simulating from a Bernoulli, and there's not really any way for it to know anything else was intended. But it looks like something the example was meant to produce something more like:

```
simulate(gm1)
#    sim_1.incidence sim_1.V2
# 1                0       14
# 2                1       11
# 3                0        9
# 4                0        5
# 5                2       20
# 6                2       16
# 7                2       19
# ...
```

Note that the following doesn't work though:

```
simulate(cbind(incidence, size - incidence)  ~ period + (1 | herd),
             newdata=simdat, family=binomial, newparams=params)[[1]]
# theta parameter vector not named: assuming same order as internal vector
# beta parameter vector not named: assuming same order as internal vector
#       incidence  
#  [1,]         1 0
#  [2,]         1 0
#  [3,]         0 1
#  [4,]         1 0
#  [5,]         1 0
```

This does, but it's a little bit clunky with the `simdat$size`:

```
simulate(cbind(incidence, size - incidence) ~ period + (1 | herd),
             newdata=simdat, family=binomial, newparams=params, weights=simdat$size)[[1]]
# theta parameter vector not named: assuming same order as internal vector
# beta parameter vector not named: assuming same order as internal vector
#       incidence   
#  [1,]        11  4
#  [2,]        14  1
#  [3,]        11  4
#  [4,]        14  1
[...]

PR #401, opened 2016-11-04
get weights from cbind response for in simulate.formula

pitakakariki:
A little bit messy but it gets the job done.

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -342,7 +342,7 @@ predict.merMod <- function(object, newdata=NULL, newparams=NULL,
             ## orig. offset: will be zero if there are no matches ...
             offset <- model.offset(model.frame(object))
             if (is.null(offset)) offset <- 0
-            
+
         } else {  ## new data specified
             ## evaluate new fixed effect
             RHS <- formula(substitute(~R,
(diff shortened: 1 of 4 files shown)


pitakakariki:
This is for https://github.com/lme4/lme4/issues/400

bbolker:
I hate to look a gift horse in the mouth, but ... could we have a couple lines of comments/documentation in the code about the logic used here? (Does this need any updates in the end-user documentation, or does it rather make the code do what the user would expect it to do in the first place?)

bbolker:
OK, just *one* more: could you change `binWtsHack` to something more like `userWts` ?  Not that we don't have other ugly stuff in the code ... (`grep -i hack R/*.R` reveals `reFormHack` and `evalhack` variables as well as, of course, lots of comments ... maybe I'll try to fix those ...) I promise I'll take it then ...

pitakakariki:
Ah, I saw `reFormHack` so I thought I was following naming convention!

I'm writing up some documentation for `?simulate.merMod`. New behaviour is what I would expect, but it can't hurt to tell people what we think it's meant to do.

bbolker:
thanks!

PR #409, opened 2017-01-31
glmer's formula includes fixed and random effects

dondelelcaro:
Fix a typo where fixed effects are mentioned twice.


bbolker:
thanks!

PR #458, opened 2018-03-09
make predict work with partial RE form

### LINKED ISSUE #457: mkNewReTrms trying to evaluate unused grouping factors
line 22 of `predict.R`, in `mkNewReTrms`, tries to evaluate the entire random-effects formula in the context of `newdata`.

```
library(lme4)
set.seed(101)
n = 300
xx<-c("r1", "r2", "r3", "r4", "r5")
xxx<-c("e1", "e2", "e3")
p=0.3
School = factor(sample(xxx, n, replace=TRUE), levels=xxx, ordered=FALSE)
Rank = factor(sample(xx, n, replace=TRUE), levels=xx, ordered=FALSE)
df1 <- data_frame(
  ID = as.integer(runif(n, min = 1, max = n/7)),
  xx1 = runif(n, min = 0, max = 10),
  xx2 = runif(n, min = 0, max = 10),
  xx3 = runif(n, min = 0, max = 10),
  School,
  Rank,
  yx = as.factor(rbinom(n, size = 1, prob = p))
)
df1 = df1[order(df1$ID, decreasing=FALSE),]
mm2 <- glmer(yx ~ xx1 + xx2 + xx3 + Rank +  (1 | ID) + (1 | School / Rank),
           data = df1,
     family = "binomial",control = glmerControl(calc.derivs =FALSE))
n11 <-  data.frame(School=factor("e1", levels =
levels(df1$School),ordered=FALSE),
    Rank=factor("r1", levels =levels(df1$Rank),ordered=FALSE),
 xx1=8.58, xx2=8.75, xx3=7.92)
predict(mm2, n11, type="response",re.form= ~(1 | School / Rank))
```

bbolker:
I'm doing this via pull request because I thought it might be worth a little bit of discussion.  These changes fix #457 , the case where `re.form` is partial (neither `~0` nor the full/original specification). The problem here is that the `predvars` info that we need in order to reconstruct data-dependent bases (`poly`, `ns`, etc.) may contain terms that are not specified by the user in `newdata` (and need not be). I spent a long time yesterday going down a rabbit hole using `model.frame()` with `re.form` and trying to make sure that everything about the model frame that was constructed (order of terms, order of interaction variables, etc.) was consistent with the original `predvars` spec, which turns out to be a giant headache.

In the end, even though it feels a little klugey, it turns out to be far simpler to assign new variables (containing `NA`) for any variables that are found in `predvar` but missing from `newdata` ...

```
            ## check that terms occur in same order in the formula, so that interactions
            ## will be constructed in the same order. From ?model.matrix:
            ## > In an interaction term, the variable whose levels vary fastest is
            ## > the first one to appear in the formula (and not in the term), so
            ## > in ‘~ a + b + b:a’ the interaction will have ‘a’ varying fastest.
            if (inForm(ss,as.symbol(":")) || inForm(ss,as.symbol("/"))) {
                browser()
                orig.vars <- all.vars(formula(object))
                interax <- c(extractForm(ss,as.symbol(":")),extractForm(ss,as.symbol("/")))
                ## (can't use vapply, results may be different lengths ...)
                interax.vars <- lapply(interax,all.vars)
                for (i in seq_along(interax)) {
                    ipos <- match(interax.vars[[i]],orig.vars)
                    if (is.unsorted(ipos)) {
                        ## if vars don't appear in same order as in original formula (e.g.
                        ## because of fixed-effect terms earlier in the formula) ... add
                        ##  those vars (in correct order) to the beginning of the formula
                        ##  to make sure interactions are constructed correctly
                        ss <- as.formula(replaceForm(ss,interax[[i]],
                                          makeOp(interax[[i]][[3]],interax[[i]][[2]],
                                                 op=head(interax[[i]]))
                                          ))
                    }
                }
            }
```

and:

```
          ff <- findbars(re.form)
            of <- findbars(formula(object))
            mm <- match(ff,of)
            if (any(is.na(mm))) stop("RE terms in re.form not found in original formula: ",deparse(ff[is.na(mm)]))
            ss <- formula(makeOp(sumTerms(ff[mm]),as.symbol("~")))
            rf2 <- terms(ss,data=newdata.NA)

```


bbolker:
decided to fix some more RE prediction issues while I was in there. Added slightly to code complexity, but hopefully it's not too bad ...

PR #507, opened 2019-02-26
The example for ranef was not working. …

brucebcampbell:
Added condVar = TRUE so the conditional variances are available for the plot.


bbolker:
thanks!

bbolker:
wait a minute: `condvar=TRUE` should be true by default now ... ? i.e., this shouldn't be necessary?

PR #525, opened 2019-06-11
models with very large numbers of random effects can cause predict to run out of memory

ifellows:
Minor change to make predict faster and avoid out of memory errors when very large numbers of random effects are in the model.

I ran into out of memory issues dealing with random intercept model where the grouping factor had hundreds of thousands of levels.

The variance structure from ranef does not appear to be used here.


ifellows:
@bbolker or @mmaechler : Thoughts on this change?

bbolker:
Sorry to let this slip through the cracks.  This makes perfect sense to me.

bbolker:
thanks!

Issue #410, opened 2017-02-03
missDataFun is deparsing entire data frames


pitakakariki:
This can get pretty slow if the data frame is big enough.

The offending line is:

    if (grepl("^\\.\\.[0-9]+$",safeDeparse(ex)))

Would it still work if it were something like one of these?

    if (!is.data.frame(ex) && grepl("^\\.\\.[0-9]+$",safeDeparse(ex)))
    # or
    if (is.symbol(ex) && grepl("^\\.\\.[0-9]+$",safeDeparse(ex)))

Some code that triggers the problem:

```
X <- expand.grid(x1=1:10, x2=1:10, x3=1:10, x4=1:10, g=letters[1:20])

X$y <- X$x1 + rnorm(10)[X$g] + rnorm(200000)
system.time(lmer(y ~ x1 + x2 + x3 + x4 + (1|g), data=X, control=lmerControl(optimizer=NULL)))

f <- function(X) {

    X$y <- X$x1 + rnorm(10)[X$g] + rnorm(200000)

    lmer(y ~ x1 + x2 + x3 + x4 + (1|g), data=X, control=lmerControl(optimizer=NULL))
}

system.time(f(X))

g <- function(X) {

    X$y <- X$x1 + rnorm(10)[X$g] + rnorm(200000)

    lme4:::missDataFun(X)
}

system.time(g(X))
```

wahani:
I believe we (@mgross) have the same issue. Especially with larger data frames this is a problem. Please find below a short reproducible example. We also provide a workaround, which is to reset the environment of the formula passed to lmer:

```r
library(lme4)

lmerTest <- function(f){
  #create simple data set
  n_u <- 1000
  n_y <- 100
  u <- rep(rnorm(n_u), each=n_y)
  id <- rep(1:n_u, each=n_y)
  x <- rnorm(n_u*n_y)
  y <- 1+ 0.5 * x + u + rnorm(n_u*n_y)
  data <- data.frame(y=y, x=x, id=id)
  #slow
  Test1 <- system.time({lmer(f, data = data)})[3]
  #workaround
  Test2 <- system.time({
    environment(f) <- list2env(data)
    lmer(f)
  })[3]
  return(list(times = c(Test1, Test2), data = data))
}

f <- y ~ x+(1|id)
lmerTestExec <- lmerTest(f = f)
lmerTestExec$times
```

We traced the problem back to the same line as already mentioned by @pitakakariki.

bbolker:
[...]

Issue #369, opened 2016-03-11
glmer error when using with lapply()


atyre2:
The error arises when running glmer inside lapply().

    library(lme4)

    ## Loading required package: Matrix

    cbpp$obs <- 1:nrow(cbpp)
    (gm2 <- glmer(cbind(incidence, size - incidence) ~ period +
        (1 | herd) +  (1|obs),
                  family = binomial, data = cbpp))

    ## Generalized linear mixed model fit by maximum likelihood (Laplace
    ##   Approximation) [glmerMod]
    ##  Family: binomial  ( logit )
    ## Formula: cbind(incidence, size - incidence) ~ period + (1 | herd) + (1 |  
    ##     obs)
    ##    Data: cbpp
    ##      AIC      BIC   logLik deviance df.resid 
    ## 186.6383 198.7904 -87.3192 174.6383       50 
    ## Random effects:
    ##  Groups Name        Std.Dev.
    ##  obs    (Intercept) 0.8911  
    ##  herd   (Intercept) 0.1840  
    ## Number of obs: 56, groups:  obs, 56; herd, 15
    ## Fixed Effects:
    ## (Intercept)      period2      period3      period4  
    ##      -1.500       -1.226       -1.329       -1.866

    listOfFormulas <- list(
      cbind(incidence, size - incidence) ~ 1 +
        (1 | herd) +  (1|obs),
      cbind(incidence, size - incidence) ~ period +
        (1 | herd) +  (1|obs)
    )
    lapply(listOfFormulas,glmer,family=binomial,data=cbpp)

    ## Error: 'data' not found, and some variables missing from formula environment

    sessionInfo()

    ## R version 3.2.3 (2015-12-10)
    ## Platform: x86_64-w64-mingw32/x64 (64-bit)
    ## Running under: Windows 7 x64 (build 7601) Service Pack 1
    ## 
    ## locale:
    ## [1] LC_COLLATE=English_United States.1252 
    ## [2] LC_CTYPE=English_United States.1252   
    ## [3] LC_MONETARY=English_United States.1252
    ## [4] LC_NUMERIC=C                          
    ## [5] LC_TIME=English_United States.1252    
    ## 
    ## attached base packages:
    ## [1] stats     graphics  grDevices utils     datasets  methods   base     
    ## 
    ## other attached packages:
    ## [1] lme4_1.1-11  Matrix_1.2-3
    ## 
    ## loaded via a namespace (and not attached):
[...]

PR #587, opened 2020-08-25
Fix data validation in corner cases

lionel-:
Hello,

I'm working on a new release of magrittr with a C implementation and I found an issue in our revdepchecks that boils down to this:

```r
library(magrittr)

# Works fine
mtcars %>% lme4::lmer("disp ~ (1 | cyl)", data = .) %>% class()
#> [1] "lmerMod"
#> attr(,"package")
#> [1] "lme4"

# Fails with a wrapper
fn <- function(data) {
  lme4::lmer("disp ~ (1 | cyl)", data = data)
}
mtcars %>% fn()
#> Error in exists(v, envir = env, inherits = FALSE) :
#>   use of NULL environment is defunct
```

A pure base R reprex can be constructed with `do.call()` (a wrapper for the C-level `Rf_eval()` function) and an artificial environment that doesn't appear in the inspectable frames:

```r
e <- new.env()
e$. <- mtcars

# Works
class(do.call(lme4::lmer, list("disp ~ (1 | cyl)", quote(.)), envir = e))
#> [1] "lmerMod"
#> attr(,"package")
#> [1] "lme4"

# Fails with a wrapper
fn <- function(data) {
  lme4::lmer("disp ~ (1 | cyl)", data = data)
}
do.call(fn, list(quote(.)), envir = e)
#> Error in exists(v, envir = env, inherits = FALSE) :
#>   use of NULL environment is defunct
```

This error comes from `missDataFun()` erroneously inferring that the data is missing and can't be evaluated. Looking at its implementation, it seems a bit brittle. This helper has caused other issues in the past:

- Wrong inference within `lapply()` (#369), which is the same problem.
- Performance issue with large data frames (#410).

Unless I'm missing something, the only purpose for this function is to slightly improve error messages when the data doesn't exist. Perhaps if it isn't worth the trouble? This PR is a suggested fix to solve the problem with magrittr, `do.call()`, and probably other corner cases. Here is the impact on error messages:

```r
# Before
lme4::lmer(z ~ 1|f, ddd)
#> Error: 'data' not found, and some variables missing from formula environment

# After
lme4::lmer(z ~ 1|f, ddd)
#> Error: bad 'data': object 'ddd' not found
```

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -895,41 +895,13 @@ checkArgs <- function(type,...) {
 ## if #3 is true *and* the user is doing something tricky with nested functions,
 ## this may fail ...
 
-## from http://stackoverflow.com/questions/14945274/determine-whether-evaluation-of-an-argument-will-fail-due-to-non-existence
-missDataFun <- function(d) {
-    ## test here
-    ff <- sys.frames()
-    ex <- substitute(d)
-    ii <- rev(seq_along(ff))
-    foundAnon <- FALSE
-    for(i in ii) {
-        ## trying to find where the original symbol is defined ...
-        ex <- eval(substitute(substitute(x, env=sys.frames()[[n]]),
-                              env = list(x = ex, n=i)))
-        if (is.symbol(ex) && grepl("^\\.\\.[0-9]+$",safeDeparse(ex))) {
-            ## testing for the dreaded "..1", "..2" pattern; this means
-            ## we are stuck in an anonymous function somewhere ...
-            ## won't be able to see whether data exist or not,
-            ## but we should not flag them as missing -- definitely
-            ## causes false positives
-            ## might be better to check against quote(..1), quote(..2),
-            ## ... but ugh ...
-            foundAnon <- TRUE
-            break
-        }
-    }
-    return(!foundAnon && is.symbol(ex) && !exists(deparse(ex)))
-}
-
 ## try to diagnose missing/bad data
 checkFormulaData <- function(formula, data, checkLHS=TRUE,
                              checkData=TRUE, debug=FALSE) {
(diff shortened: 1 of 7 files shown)


R/utilities.R
```diff
@@ -895,41 +895,13 @@ checkArgs <- function(type,...) {
...
-
 ## try to diagnose missing/bad data
 checkFormulaData <- function(formula, data, checkLHS=TRUE,
                              checkData=TRUE, debug=FALSE) {
-    nonexist.data <- missDataFun(data)
-    wd <- tryCatch(eval(data), error = identity)
-    if (wrong.data <- inherits(wd,"simpleError")) {
-        wrong.data.msg <- wd$message
+    wd <- tryCatch(force(data), error = identity)
+    if (bad.data <- inherits(wd,"simpleError")) {
```
lionel-, line 902: Should probably be `inherits(wd, "error")` actually.

bbolker:
The existing code is definitely fragile and ugly, and I appreciate the contribution.  On other hand, changing it makes me nervous because we have had lots of trouble with these issues in the past, and there are lots of different ways to call the function, some of which could easily be missing from our test cases.  On the other hand, if it passes all of the test cases we have *and* doesn't break anything in our own reverse-dependency checks it would be nice to clean up this corner.

bbolker:
Just checked and this is still OK with the timing comparisons given for #410 [here](https://github.com/lme4/lme4/issues/410#issuecomment-311092416).

bbolker:
oops, accidentally pushed some other crap in https://github.com/lme4/lme4/pull/587/commits/4949036cf52f88614ed437f77b85524b5ed30361 but I think it's harmless ...

PR #629, opened 2021-06-10
Remove vestigial object reference

MichaelChirico:
Ben mentioned an overhaul to the vignette; removing this line in the meantime

https://stat.ethz.ch/pipermail/r-sig-mixed-models/2021q2/029554.html


bbolker:
thanks!

Issue #631, opened 2021-06-15
"object is not a matrix" error in `predict()`


strengejacke:
``` r
library(lme4)
data <- read.csv("data_(1).csv")
model <- lmer(response ~ condition_bystanders + (condition_bystanders|ID), 
              data = data)

d <- as.data.frame(expand.grid(unique(data$condition_bystanders), unique(data$ID)))
colnames(d) <- c("condition_bystanders", "ID")
predict(model, newdata = d)
#> Error in model.matrix.default(eval(substitute(~foo, list(foo = x[[2]]))), : object is not a matrix
```

<sup>Created on 2021-06-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

_Originally posted by @strengejacke in https://github.com/strengejacke/ggeffects/issues/221#issuecomment-861855638_

[data_(1).zip](https://github.com/lme4/lme4/files/6658928/data_.1.zip)

bbolker:
Weird. I assume it has to do with factor/character mismatches, but I'll take a look. Just surprising that we still have errors like this to discover (at a first glance it's not caused by recent changes, but I'll check that too).

bbolker:
Same error occurs for 1.1-26 and 1.1-25, and with `options(stringsAsFactors = TRUE)` set ... ???

strengejacke:
But this is not related to the discussion at https://github.com/glmmTMB/glmmTMB/issues/602? I know it's a different issue in glmmTMB, but you mentioned the code that checked factor levels/order here: https://github.com/glmmTMB/glmmTMB/issues/602#issuecomment-651194250

bbolker:
Don't know, I will check. Same error if I use `d2 <- as.data.frame(lapply(d, as.character))` as `newdata` (so we shouldn't have a problem with factor levels ??)

bbolker:
This seems bizarre.  All I can see so far is that at some point one of the factor gets a contrast attribute that is a `Matrix` rather than a `matrix` - I think that's probably the problem, but I don't yet know where/why it's happening.

bbolker:
Only happens if the number of levels of the grouping variable is > 100

Problem has to do with [this](https://github.com/lme4/lme4/blob/67afecbff4831de9caf6b5891e8f307a01f7197e/R/predict.R#L159), [this](https://github.com/lme4/lme4/blob/67afecbff4831de9caf6b5891e8f307a01f7197e/R/predict.R#L194-L195).

[...]

PR #632, opened 2021-06-16
Convert sparse contrast matrices in predict

bwiernik:
Closes https://github.com/lme4/lme4/issues/631


bbolker:
Closing this as I think there is a better fix in master. Feel free to re-open (thanks for the effort!)

bwiernik:
No worries at all! I was mostly curious as to why it was failing myself.

PR #642, opened 2021-09-02
Update minimum version of Rcpp

### LINKED ISSUE #641: Rcpp version in DESCRIPTION is too old
I've just run into an issue with a missing function in Rcpp because the version required by lme4 is older than is actually supported. `Rcpp_precious_remove` wasn't introduced until Rcpp 1.0.6, so I would suggest updating the required version in DESCRIPTION at least to that (or to 1.0.7 if you're not planning to test lme4 with the previous release).

NikNakk:
Fixes #641 by increasing minimum required Rcpp version to the current release (1.0.7)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -31,7 +31,7 @@ Description: Fit linear and generalized linear mixed-effects models.
     methods.  The core computational algorithms are implemented using the
     'Eigen' C++ library for numerical linear algebra and 'RcppEigen' "glue".
 Depends: R (>= 3.2.0), Matrix (>= 1.2-1), methods, stats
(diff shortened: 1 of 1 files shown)


bbolker:
This might work, but it doesn't really address the root problem (see #641); it will be better to implement something that checks binary versions and gives an appropriate warning.  In particular, if one has a version of `RcppEigen` installed that was built with an older version of `Rcpp`, I don't think that bumping the `Rcpp` requirement for `lme4` will necessarily solve the problem. (What's needed is a reinstall-from-source of `RcppEigen` with the matching `Rcpp` version.)

PR #662, opened 2022-02-14
use skip_if_not_installed for optional tests

MichaelChirico:


mmaechler:
What problem does this fix?   To me it only engrains the dependency on testthat  (rather than using base R)

MichaelChirico:
this is a testthat suite so I don't quite follow why we wouldn't want to make things more canonical for that framework.

I wouldn't say it solves a problem, more that it offers a benefit -- users running the suite without the package installed are served with a nice summary of skipped tests after test_check() completes.

currently, you might have no idea some tests were skipped unless you read the test scripts carefully; with proper skip(), this potential issue is surfaced alongside the rest of the test summary.

MichaelChirico:
I suppose alternatively we could add an else branch to the requires that prints something using only base functionality, WDYT?

bbolker:
I had been considering a move to `tinytest`: https://www.avrahamadler.com/2022/01/20/reduce-dependency-hell/

MichaelChirico:
noted; feel free to close in that case

PR #674, opened 2022-03-30
Update URL for `BLUPS are not best`

nfultz:
I was reading the code for bootstrapping and found that one of the links didn't work anymore, so fixing it while I am here.


bbolker:
thanks! (I love this aspect of public software development ...)

PR #690, opened 2022-07-09
fix test for optimx presence

MichaelChirico:
Dunno what I was thinking in fbdb5360ea0176cfbbe7ef341bcf94c14b149ef1 😄

Also include an opt-out for the tests in tests/boundary.R also requiring optimx

### DIFF
--- a/tests/boundary.R
+++ b/tests/boundary.R
@@ -65,7 +65,7 @@ if (.Platform$OS.type != "windows") {
     stopifnot(all.equal(unname(getME(fm5c,"theta")), 0.21067645, tolerance = 1e-7))
 					#	 0.21067644264 [64-bit, lynne]
 
-    ##{
+    if (require("optimx")) {
     ## additional stuff for diagnosing Nelder-Mead problems.
 
     library(optimx)
@@ -90,7 +90,7 @@ if (.Platform$OS.type != "windows") {
                  loglik = sapply(mList, logLik))
     res
     print(sessionInfo(), locale=FALSE)
-    ##}
+    }
 
 ######################
     library(lattice)
--- a/tests/boundary.R
+++ b/tests/boundary.R
@@ -68,7 +68,6 @@ if (.Platform$OS.type != "windows") {
     if (require("optimx")) {
     ## additional stuff for diagnosing Nelder-Mead problems.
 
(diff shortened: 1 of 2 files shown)


bbolker:
Can you say briefly what the thinko/logic error was in the original patch?
Unfortunately, I very recently pushed a new version of the pkg to CRAN. I'm happy to merge this, but is there a reason to push a fix to CRAN in the near future, or can we just let it wait for the next release?

MichaelChirico:
the whole issue is that when run on a  machine without optimx, the tests edited here will fail.

but with ||, that test fails and we still try packageVersion of a package that doesn't exist. what we want is to pass requireNamepace first before trying packageVersion

this won't happen on CRAN because optimx is in Suggests and CRAN always runs with your Suggests available, so there's no rush for CRAN release.

bbolker:
BDR had actually set up a platform for checking packages with suggested packages missing. I think I got hassled about such failures in one of my packages (see [here](https://stat.ethz.ch/pipermail/r-package-devel/2022q2/008036.html)). I don't see that being tested any more - perhaps CRAN got too much pushback?

Do you know offhand if BDR's auxiliary test platforms are listed/documented somewhere? The CRAN platforms are listed [here](https://cran.r-project.org/web/checks/check_flavors.html)  (Update, they can be found [cryptically] [here](https://www.stats.ox.ac.uk/pub/bdr/?C=M;O=D))

MichaelChirico:
such a test is slightly different, IIUC. that would trigger if e.g. you use A in your tests but don't list it in your Suggests. I'm not sure if that's still done (presumably codetools could detect such an issue statically, anyway, so R CMD check should report it)

lme4 is definitely obeying CRAN rules -- this PR is only a convenience feature for passing more tests on systems without the full set of dependencies available

bbolker:
I don't think so. My point is that BDR **was** testing on systems without weak dependencies available (and, at least for a little while, CRAN was flagging packages that failed in that context - see the mailing list thread I linked above). Anyway, all was previously quiet on the CRAN front and hopefully will continue to be so.

MichaelChirico:
I see... that's rather strict!

I have been making small PRs like this to _many_ packages so I doubt CRAN is doing anything like that recently

PR #746, opened 2023-10-06
Header clean-up to fix #745

### LINKED ISSUE #745: use headers from Matrix, not RcppEigen
I was quite surprised to learn that **RcppEigen** contains static, now somewhat stale copies of the **Matrix** headers, and that those copies are included in `RcppEigen.h`.  A corollary is that the following list of the reverse LinkingTo of **Matrix**,

```
> rrLT <- function(p) unique(unlist(tools::package_dependencies(p, which = "LinkingTo", reverse = TRUE, recursive = TRUE), FALSE, FALSE))
> length(p1 <- rrLT("Matrix"))
[1] 36
```

is incomplete: we actually need to look at a much longer list,

```
> length(p2 <- rrLT(c("Matrix", "RcppEigen")))
[1] 449
```

when testing the effect of changes to the Matrix C API in package space.  A static analysis of the packages in `setdiff(p2, p1)` reveals that **lme4** is the _only_ package making use of the **RcppEigen** copies of the headers.  So I am creating this issue partly to ask: why?

It seems wrong to me that **Matrix** should have these "hidden" 413 reverse LinkingTo.  Shouldn't **lme4** change to accessing CHOLMOD via **Matrix** instead of **RcppEigen**?  Then eventually **RcppEigen** could stop packaging the headers, which have been liable to become stale and have created this weird situation beneficial (if you can call it that) seemingly only to **lme4**.

Edit: Well, I see now that the headers are there to support Eigen's CHOLMOD module.  It is possible that packages are using the module without calling any of the functions declared in the copies of our headers.  Still, I think that the situation can be improved ...

jaganmn:
Note that these changes are entirely conditional on RcppCore/RcppEigen#131, hence `DESCRIPTION` was modified to depend on a patched **RcppEigen** and CRAN submission must be co-ordinated with @eddelbuettel.

Edit: Now I see that **RcppEigen** has `R (>= 3.6.0)` whereas **lme4** has `R (>= 3.5.0)`.  Do we know why **RcppEigen** does not support 3.5.0?

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -31,7 +31,8 @@ cache/.*
 README.md
 TODO.md
 R/TAGS
-src/mcmcsamp.cpp
+src/mcmcsamp\.h$
+src/mcmcsamp\.cpp$
 ^\.github$
 revdep/
 Rplots.pdf
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,7 @@ Description: Fit linear and generalized linear mixed-effects models.
     methods.  The core computational algorithms are implemented using the
     'Eigen' C++ library for numerical linear algebra and 'RcppEigen' "glue".
 Depends: R (>= 3.5.0), Matrix (>= 1.2-1), methods, stats
-LinkingTo: Rcpp (>= 0.10.5), RcppEigen
+LinkingTo: Rcpp (>= 0.10.5), RcppEigen (>= 0.3.3.9.3.2), Matrix
 Imports: graphics, grid, splines, utils, parallel, MASS, lattice, boot,
     nlme (>= 3.1-123), minqa (>= 1.1.15), nloptr (>= 1.0.4)
 Suggests:
--- a/src/lme4CholmodDecomposition.h
+++ b/src/lme4CholmodDecomposition.h
@@ -10,6 +10,7 @@
 #define LME4_CHOLMODDECOMPOSITION_H
 
 #include <RcppEigen.h>
+#include <RcppEigenCholmod.h>
 
 namespace lme4 {
 
@@ -42,8 +43,8 @@ namespace lme4 {
                 // viewAsCholmod(matrix.template selfadjointView<_UpLo>()) :
                 viewAsCholmod(matrix);
 
-            cholmod_factorize_p(&A, &beta, fset.data(), fset.size(),
-                                factor(), &cholmod());
+            R_MATRIX_CHOLMOD(factorize_p)(
+                &A, &beta, fset.data(), fset.size(), factor(), &cholmod());
 
             this->m_info = Eigen::Success;
             m_factorizationIsOk = true;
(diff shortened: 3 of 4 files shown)


jaganmn:
It still seems to me that these templates,

https://github.com/jaganmn/lme4/blob/rcppeigen_cholmod/src/lme4CholmodDecomposition.h#L75-L120

are redundant given that class `lme4::lme4CholmodDecomposition` inherits from `Eigen::CholmodDecomposition` and **RcppEigen** already defines templates for the parent class,

https://github.com/jaganmn/RcppEigen/blob/rcppeigen_cholmod/inst/include/RcppEigenCholmod.h#L35-L82

Can't the templates for the parent be inherited ... ? Will take another look tomorrow ...

eddelbuettel:
> Do we know why RcppEigen does not support 3.5.0?

The ChangeLog entry (from the `git blame` date) reminded me -- t's because of FCONE and USE_FC_LEN_T

https://github.com/RcppCore/RcppEigen/blob/56dc0cca08fdc796a9402ba0803d58e924ed7f67/ChangeLog#L51-L55

jaganmn:
> > Do we know why RcppEigen does not support 3.5.0?
> 
> The ChangeLog entry (from the `git blame` date) reminded me -- t's because of FCONE and USE_FC_LEN_T
> 
> https://github.com/RcppCore/RcppEigen/blob/56dc0cca08fdc796a9402ba0803d58e924ed7f67/ChangeLog#L51-L55

Great - then we can definitely get **RcppEigen** to support R 3.5.0 by following the advice in WRE, [here](https://stat.ethz.ch/R-manual/R-patched/doc/manual/R-exts.html#Fortran-character-strings), as **Matrix** already does (see `Matrix/src/Lapack-etc.h`, included by, e.g., `Matrix/src/kappa.c`).  That would save **lme4** some headache.  Shall I add to the existing PR or create a new one?

eddelbuettel:
Maintainer speaking here:  Why would I care about R 3.5.0?  We are now at R 4.3.1.

jaganmn:
> Maintainer speaking here: Why would I care about R 3.5.0? We are now at R 4.3.1. (I am not opposed, I just don;t see it being worth anything tangible. As usual, if this is an itch you really must scratch file a PR and demonstrate that it does no harm.)

Thanks.  I don't care, personally.  I guess if @bbolker and @mmaechler (a) approve this PR subject to rev. dep. checking and (b) would want to merge without preventing users of 3.5.0 <= R < 3.6.0 from installing future versions of **lme4**, then they can say so, and only in that case would I file the additional PR with **RcppEigen**.

bbolker:
*all other things being equal* I like to maintain backward compatibility, but ... I think it's probably OK to deprecate/break back-compatibility with 3.5.0, which is 5+ years old now (e.g. [Ubuntu LTS](https://ubuntu.com/about/release-cycle) and  [Debian LTS](https://wiki.debian.org/LTS) only provide a 5-year window ...)

bbolker:
Note to self: before PR is merged we need to follow [these instructions](https://stackoverflow.com/questions/75435418/r-install-github-how-to-install-pull-request) to install from [this PR](https://github.com/RcppCore/RcppEigen/pull/131), i.e.

```
remotes::install_github("RcppCore/RcppEigen", ref = remotes::github_pull(131))\
```

bbolker:
Another note to self: this has been reverted temporarily; to restore should `git cherry-pick d66266f77304` (and possibly also cherry-pick `6fbb956a60dd`, which bumps the required version of `RcppEigen`)

PR #755, opened 2023-11-06
Fixes #753

### LINKED ISSUE #753: `predict()` fails when `se.fit = TRUE` and `re.form = NULL`
Not sure if this is a bug, or a mis-specification from my side?

``` r
library(lme4)
#> Loading required package: Matrix

m1 <- lmer(Reaction ~ Days + (1 + Days | Subject), data = sleepstudy)
d <- as.data.frame(expand.grid(list(Days = 0:9, Subject = 308)))

predict(m1, newdata = d, se.fit = TRUE, re.form = NULL, allow.new.levels = TRUE)
#> Error in h(simpleError(msg, call)): error in evaluating the argument 'x' in selecting a method for function 'rowSums': Cholmod error 'A and B inner dimensions must match' at file ../MatrixOps/cholmod_ssmult.c, line 80
predict(m1, newdata = d, se.fit = TRUE, re.form = NULL, allow.new.levels = FALSE)
#> Error in h(simpleError(msg, call)): error in evaluating the argument 'x' in selecting a method for function 'rowSums': Cholmod error 'A and B inner dimensions must match' at file ../MatrixOps/cholmod_ssmult.c, line 80
predict(m1, newdata = d, se.fit = FALSE, re.form = NULL, allow.new.levels = TRUE)
#>        1        2        3        4        5        6        7        8 
#> 253.6637 273.3299 292.9962 312.6624 332.3287 351.9950 371.6612 391.3275 
#>        9       10 
#> 410.9937 430.6600
predict(m1, newdata = d, se.fit = FALSE, re.form = NULL, allow.new.levels = FALSE)
#>        1        2        3        4        5        6        7        8 
#> 253.6637 273.3299 292.9962 312.6624 332.3287 351.9950 371.6612 391.3275 
#>        9       10 
#> 410.9937 430.6600
```

<sup>Created on 2023-11-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

emitanaka:
Now `allow.new.levels = TRUE` works (by making Z = 0)

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -561,7 +561,11 @@ predict.merMod <- function(object, newdata=NULL, newparams=NULL,
         }
     } else {
         if(isRE(re.form)) {
-            Z <- t(newRE$Zt)
+            Zoriginal <- getME(object, "Z")
+            Znew <- t(newRE$Zt)
+            Z <- Matrix(0, nrow = nrow(X), ncol = ncol(L))
+            # FIXME: this will likely trip up for complex forms of re.form
+            Z[, which(colnames(Zoriginal) %in% colnames(Znew))] <- Znew
         } else {
             ## this is inefficient and we could just calculate 
             ## X %*% Cmat[X part only] t(X) instead
--- a/R/predict.R
+++ b/R/predict.R
@@ -565,7 +565,10 @@ predict.merMod <- function(object, newdata=NULL, newparams=NULL,
             Znew <- t(newRE$Zt)
             Z <- Matrix(0, nrow = nrow(X), ncol = ncol(L))
             # FIXME: this will likely trip up for complex forms of re.form
-            Z[, which(colnames(Zoriginal) %in% colnames(Znew))] <- Znew
+            ind <- which(colnames(Zoriginal) %in% colnames(Znew))
+            if(length(ind)) {
+              Z[, ind] <- Znew
+            }
         } else {
             ## this is inefficient and we could just calculate 
             ## X %*% Cmat[X part only] t(X) instead


bbolker:
Hmm.  I haven't looked at this in detail, but I'm a little bit puzzled by the need to do complex (and/or fragile) subsetting of the original Z matrix. The [mkNewReTrms](https://github.com/lme4/lme4/blob/ce9bd2d9cd5ef2945ccd6be2455d5f1b05670b58/R/predict.R#L158-L292) function is admittedly too long and needs refactoring, but it's intended to isolate all of the complex matching ... in particular, the internally defined [get_re function](https://github.com/lme4/lme4/blob/ce9bd2d9cd5ef2945ccd6be2455d5f1b05670b58/R/predict.R#L271-L284) is supposed to pick out the BLUP values corresponding to the random effects that are to be included in the prediction ...  can we see a reason this machinery isn't working as intended?

emitanaka:
I agree that the Z subsetting is fragile. Currently, it's not working as the [`Cmat`](https://github.com/lme4/lme4/blob/master/R/predict.R#L535-L549) is based on the original $\mathbf{X}$ and $\mathbf{Z}$ and so the dimension of the new $\mathbf{Z}$ from `mkNewReTrms` doesn't match if the new data supplied only contains subset of random effects. I'll have to check what `get_re` function does, but ideally, if we can get the column index of $\mathbf{Z}$ that matches with the supplied new data then that'd be a better approach I think. The other approach is to recalculate`Cmat` based on the new data but I think this is too painful?

bbolker:
Yes, I think it might not be too hard to extend `mkNewReTrms` to return a set of indices that would do what you need ...

After a brief look it's harder than I thought, but here's how I think I would approach it:

* get the `ranef()` object from the original model
* replace the numeric values with indices in an order that corresponds to the columns of Z (in other words, each data frame corresponding to a random effect term would be filled with integer values in *rowwise* order)
* we already have code (the internal `get_re` function) that picks out elements of `b` that are included in the new RE terms. We could modify/copy this framework to extract the relevant indices.

bbolker:
I think this has been resolved.

PR #793, opened 2024-05-31
Add missing Matrix version dependency

MichaelChirico:
AFAICT this macro was added in Matrix 1.6.2:

https://github.com/lme4/lme4/blob/2d82cbee952c23e2d3f8492489da7bd5a40dfb2f/src/lme4CholmodDecomposition.h#L72

It _may_ be avoidable by just copying the macro definition:

https://github.com/cran/Matrix/blob/d91a3f2149572d5e4801cb9d4c28d55ffd3642e8/inst/include/Matrix/cholmod.h#L4166-L4168

But not sure if that's hiding some other API pieces that were added at the same time.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,7 +34,7 @@ Description: Fit linear and generalized linear mixed-effects models.
     methods.  The core computational algorithms are implemented using the
     'Eigen' C++ library for numerical linear algebra and 'RcppEigen' "glue".
 Depends: R (>= 3.5.0), Matrix (>= 1.2-1), methods, stats
-LinkingTo: Rcpp (>= 0.10.5), RcppEigen (>= 0.3.3.9.4), Matrix
+LinkingTo: Rcpp (>= 0.10.5), RcppEigen (>= 0.3.3.9.4), Matrix (>= 1.6.2)
 Imports: graphics, grid, splines, utils, parallel, MASS, lattice, boot,
     nlme (>= 3.1-123), minqa (>= 1.1.15), nloptr (>= 1.0.4)
 Suggests:


bbolker:
I wouldn't want to depend on Matrix 1.7 (since that would lock us into R >= 4.4), but Matrix 1.6.2 only requires R >= 3.5.0 which seems fine at this point ... thanks.

jaganmn:
Sorry, but did someone encounter a build failure suggesting that a dependency on **Matrix** 1.6-2 was necessary?  When I patched **RcppEigen** (to become version 0.3.4.0.0) and **lme4** (to become version 1.35-1.1), I remember being careful to preserve compatibility with earlier versions of **Matrix** not defining `R_MATRIX_CHOLMOD`.  The commits are here: [**lme4**](https://github.com/RcppCore/RcppEigen/commit/652cb9945066745cfd595128d5c6042d045335d2) [**RcppEigen**](https://github.com/lme4/lme4/commit/d66266f773041f7cb3f37cf71d7450565a50ded4)

Notably:

* **lme4** >= 1.35-1.1 depends on **RcppEigen** >= 0.3.4.0.0
* **lme4** >= 1.35-1.1 has a `lme4CholmodDecomposition.h` that includes `RcppEigenCholmod.h`
* **RcppEigen** >= 0.3.4.0.0 has an `RcppEigenCholmod.h` that conditions on the version of **Matrix** being linked:
  
  ```c
  #include <Matrix.h>
  #ifndef R_MATRIX_CHOLMOD /* Matrix <= 1.6-1.1 */
  # define R_MATRIX_CHOLMOD(_NAME_) M_cholmod_ ## _NAME_
  # define M_cholmod_start M_R_cholmod_start /* sigh */
  #endif


  #include <Eigen/CholmodSupport>
  ```
  
  https://github.com/RcppCore/RcppEigen/blob/91b98157aa60fd300ba327aebfb6d5d8462fba66/inst/include/RcppEigenCholmod.h#L27-L33

jaganmn:
The following succeeds for me ...

```bash
$ R CMD INSTALL Matrix_1.6-1.1.tar.gz
$ R CMD INSTALL lme4_1.1-35.1.tar.gz
```

MichaelChirico:
Hi @jaganmn, I am trying to update lme4 1.1-34 --> 1.1-35.3. We have Matrix 1.5-4.1, RcppEigen 0.3.3.7.0 (doesn't meet the requirement 0.3.3.9.4, but I'm having trouble updating that package too & don't think it matters for the below).

Here's the corresponding `R_MATRIX_CHOLMOD()` usage:

https://github.com/cran/lme4/blob/298149487759297ade87ae41c44c7f0a5d3a7458/src/lme4CholmodDecomposition.h#L47

And DESCRIPTION not requiring a high enough Matrix version (AFAICT):

https://github.com/cran/lme4/blob/298149487759297ade87ae41c44c7f0a5d3a7458/DESCRIPTION#L36-L37

It's of course possible I'm missing something, or that it can be chalked up to our highly idiosyncratic setup for R packages. But at a glance I still think it's correct (for the lme4 version in question, at least).

bbolker:
I was able to install `lme4` 1.1-35.3 with Matrix 1.5-1 (in `rocker/r-ver:4.3.1`) and run some basic models and tests.  Can you say specifically what's failing when you try to install `lme4`  1.1-35.3 with Matrix 1.5-4.1 installed?

MichaelChirico:
Here's the compilation error:

```
lme4/src/lme4CholmodDecomposition.h:47:30: error: reference to non-static member function must be called
   47 |             R_MATRIX_CHOLMOD(factorize_p)(
      |                              ^~~~~~~~~~~
lme4/src/lme4CholmodDecomposition.h:64:52: error: use of undeclared identifier 'solve'
   64 |             cholmod_dense* x_cd = R_MATRIX_CHOLMOD(solve)(
      |                                                    ^
lme4/src/lme4CholmodDecomposition.h:72:30: error: use of undeclared identifier 'free_dense'
   72 |             R_MATRIX_CHOLMOD(free_dense)(&x_cd, &cholmod());
      |                              ^
lme4/src/predModule.cpp:9:10: fatal error: 'RcppEigenStubs.cpp' file not found
    9 | #include <RcppEigenStubs.cpp>
      |          ^~~~~~~~~~~~~~~~~~~~
4 errors generated.
```

The last one is from the RcppEigen version mismatch. AFAICT updating RcppEigen won't make the other part go away though?

MichaelChirico:
I see, maybe it will make the other part go away after all :)

https://github.com/cran/RcppEigen/blob/4e0fdf6bf1e0f3d36a984ad32720c157d3a94c7d/inst/include/RcppEigenCholmod.h#L28-L31

jaganmn:
Sorry, I made a typo in https://github.com/lme4/lme4/pull/793#issuecomment-2183427685, which I've just fixed with an edit.  I said that the critical **RcppEigen** version was 0.3.4.0.0 but it was actually 0.3.3.9.4.  Hence the minimum **RcppEigen** version should _not_ need to increase from 0.3.3.9.4 to 0.3.4.0.0.

MichaelChirico:
Sorry for the mixup, and thanks for the extra care to keep things back-compatible @jaganmn!

bbolker:
OK. I will definitely back out the 1.6-2 requirement; don't know if I will bother doing the archaeology to see if it can go all the way back to 1.2-1 (which was from 2015)?

jaganmn:
You depend on R 3.5, which was bundled with **Matrix** 1.2-14.  That would bring you to 2018, I guess.

bbolker:
Matrix 1.4.1 (2022-03-23) + lme4 1.1.35.3 + R 4.3.1 passes R CMD check. I didn't manage to any earlier versions in the rocker/r-ver:4.3.1 container. Anything earlier fails with "dgeMatrix.c:570:20: error: ‘DOUBLE_EPS’ undeclared". I'm sure we could in principle construct a (Matrix version × R version × lme4 version) table (i.e. earlier versions of Matrix might work with *earlier* versions of R), but I don't think I want to work that hard. Options:

1. change Matrix requirement to 1.4.1, which we at least know works with modern lme4 and modern(ish) R.
2. leave the Matrix requirement at 1.2-14 and wait for someone to complain that it doesn't actually work.

jaganmn:
I'll install R 3.5 later tonight and let you know.

jaganmn:
TL;DR: **lme4** could use something like `R (>= 3.6)` and `Matrix (>= 1.2-3)`.

**lme4** depends on **RcppEigen** >= 0.3.3.9.4, which depends on R >= 3.6, so I configured R 3.6.0 in a Debian 10 container (`docker.io/library/debian:buster`) with:

```
$ svn checkout -r 76424 https://svn.r-project.org/R/branches/R-3-6-branch R
$ cd R && ./tools/rsync-recommended && ./configure
```

I installed the latest versions of all recursive dependencies of **lme4** except for **MatrixModels**, **evaluate**, **gsl**, **lavaan**, **lme4**, **mvtnorm**, **pbkrtest**, and **phylolm** whose latest versions depend on R > 3.6 (**MatrixModels** and **lme4** depend on a version of **Matrix** newer than the one in `src/contrib/3.6.0/Recommended/`, **lavaan** is broken due to https://github.com/yrosseel/lavaan/issues/360, **mvtnorm** is broken due to backwards incompatible LAPACK calls).  I got these versions from the CRAN archive:

```
MatrixModels     evaluate          gsl       lavaan         lme4      mvtnorm 
     "0.5.0"       "0.23"      "2.1.7"     "0.6.17"   "1.1.35.3"      "1.1.3" 
    pbkrtest      phylolm 
     "0.5.1"        "2.6" 
```

**Matrix** 1.2-0, 1.2-1, and 1.2-2 fail to install under R 3.6.0 with:

```
** byte-compile and prepare package for lazy loading
Creating a generic function for 'chol2inv' from 'base' in package 'Matrix'
    (from the saved implicit definition)
Creating a generic function for 'chol2inv' from package 'base' in package 'Matrix'
Error in match.call(definition, call, expand.dots, envir) : 
  unused argument (mode = c("missing", "methods"))
Error: unable to load R code in package 'Matrix'
Execution halted
ERROR: lazy loading failed for package 'Matrix'
```

**lme4** 1.1-35.4 (with strict **Matrix** dependency reverted) checks OK under **Matrix** 1.2-3 and 1.2-4, which install fine.

FWIW, I adapted an old script to get a list of dependencies whose latest versions do not support the current R version, though it has no way to detect the broken **lavaan** and **mvtnorm**:

```r
options(repos = "https://cran.r-project.org", pkgType = "source")

f <- c("OS_type", "subarch", "R_version")
ap <- available.packages(filters = f[1:3])
ap.latest <- available.packages(filters = f[1:2])

w <- c("Depends", "Imports", "LinkingTo", "Suggests", "Enhances")
d5  <- unlist(tools::package_dependencies("lme4", db = ap.latest, which = w[1:5]))
d53 <- unlist(tools::package_dependencies(    d5, db = ap.latest, which = w[1:3], recursive = TRUE))

## Packages needed for a full 'check'
d <- sort(unique(c(d5, d53)))
## Packages needed from the CRAN archive (modulo 'recommended')
d.archive <- intersect(d, setdiff(ap.latest[, "Package"], ap[, "Package"])) # MASS Matrix evaluate gsl lattice pbkrtest phylolm
```

I'll ping Torsten Hothorn about **mvtnorm** ...

bbolker:
Thanks!

Issue #209, opened 2014-06-01
release tasks for 1.1-7


bbolker:
We should push a version of 1.1-7 soon (primarily to get the scaled-gradient convergence test pushed out to the world so people will stop asking ...)

My thoughts on release-critical tasks:
- resolve the (probably linked) issues of Karl Millar's patch; valgrind issues; rectangular/square `Zt` hacks (#200)
- switch `glmer` to `optimizer="bobyqa"` by default?  (seems to be generally slightly better than the current BOBYQA/Nelder-Mead combination)

I don't know if there is other important stuff/low-hanging fruit we should try to do in the short term.
- Fixing the new `simulate()` bug, #212, definitely needs to be added to the tasks. That did work fine previously (even in 1.1-5 I think).

PR #804, opened 2024-07-02
Add argument to simulate() to allow user-specified distn cluster rand…

RossBoylan:
Allows specification of arbitrary non-Normal cluster random effects in simulation.  See `vignettes/simulation.Rmd` for motivation, design issues, and a trace of the old code.

This facility would be useful for a package I plan to release, but will also serve other goals, such as testing robustness against violations of the assumed Normal distribution of cluster random effects.  The change is transparent to existing client code.

* modify `simulate.merMod()` and `.simulateFun()` to take an optional user-specified cluster random effects generator.
* update documentation
* add unit test for new feature
* add vignette `simulation.Rmd` describing the old code and discussing the design of this modification.
* update version and add self as contributor

The unit test is an addition to `tests/simulate.R`, and so is not a `testthat` test. I think this means `devtools::test()` will not trigger it, though `devtools::check()` does. Also, the test will *not* run unless `LME4_TEST_LEVEL >= 2`, and so ordinarily it will not run--same as all the tests in `simulate.R`.

In running this I generated a new `tests/testthat/test-summary_testleve1_1.rda`. Although the file is under version control, I have not committed the changes.  It also generated a `..._testlevel_2.rda` file, which I left not under version control.

When I raise `LME4_TEST_LEVEL` to 2 it generates a lot of test failures in code I didn't touch.  However, the new test does pass.

The "vignette" I added isn't really intended to be user facing.

I have not considered how this would extend to more than a single random effect.


bbolker:
- Running now, I don't get any errors at `LME4_TEST_LEVEL=10` (AFAIK all of the test-level statements are written as inequalities, so this should encompass level 2). Can you double check?
- I'm not sure about the vignette - it seems like it might belong in `misc/notes` instead? (This is `.gitignore` but you can force it in with `git add -f`)
-  There are a few files that were flagged by GitHub as being completely changed (i.e. every line changed), probably due to some stupid whitespace thing (like the difference between DOS and Unix line separator format). That's not the end of the world, but I have a minor preference for you making a clean PR by making a clean fork and copying your changes over ...
- in the past I had to make sure to save `.rda` etc. files with `version=2` for back-compatibility, but now that we require R >= 3.6 (due to Rcpp and/or Matrix requirements) I think that might not be a thing any more ...

RossBoylan (commented):
A bit more of a version bump seems appropriate; my package needs to have a dependency that assures the simulate feature is available, and making that a dependency way out on the patch level seems odd.  I'm not even sure the R packaging dependency mechanism allows it.

bbolker:
I'm not sure what you mean.  The `simulate` S3 generic has been in R for a long time; looking at `git blame`,  the `simulate()` generic was [already available in 2005](https://github.com/r-devel/r-svn/commit/5c51a2b5205ed6f6920e23e5b0e235a56d0021f1) (R 2.2.1).

RossBoylan:
The "simulate feature" I was referring to is the new argument allowing specification of alternate random effects distributions.

bbolker:
I'm still not sure my how to interpret "my package needs to have a dependency that assures the simulate feature is available". I guess we could bump all the way to 1.2, but this seems like a relatively minor feature addition ...

RossBoylan:
The `DESCRIPTION` file for my package needs to indicate that it requires a version of `lme4` that understands the `cluster.rand` argument to `simulate()` and related functions.  Using my test version of `lme4` I have (for my package)
```
Imports: 
    methods,
    stats,
    lme4 (>= 1.2),
    cubature,
    truncnorm
```
I have not studied the `R` specification for version number dependencies enough to be sure if `lme4 (>= 1.1-35.6)` would even work, but I do know that under semantic versioning patch-level changes should not change interfaces.  Hence my thought to bump the minor version from .1 to .2.

Sorry for the delay; I was on a mini-vacation.

bbolker:
We are not disciplined enough to do proper semantic versioning (we'd probably be at release 25 by now ...).  If you can live with a bump as far as 1.1.36, that would be more consistent with what we've done historically.

bbolker:
What do you think about version 1.1.36? Can you submit a clean PR?

mmaechler (changes requested):
I'd never approve of changes that replace everything (as in this case).
Please ensure *MINIMAL* changes, i.e.,
keep everything exactly "as is" .. unless you need to change that line.

Yes, you can "blame" the software (editor/..) you used to produce the changes. ==> use a better software to edit the source files, then.

RossBoylan:
> I'd never approve of changes that replace everything (as in this case). Please ensure _MINIMAL_ changes, i.e., keep everything exactly "as is" .. unless you need to change that line.
> 
I think the change I pushed shortly before your comment brings things back to the minimal change state, and viewing this branch against master on the web in github show that too.  So I'm not sure what you're talking about.

It is true that the last commit has a bunch of whole-file whitespace changes, but that is only because it was undoing similar changes in my earlier commit in the clusterre branch.

BTW, my scan for files with DOS line endings for master (i.e., untouched by me) does show a number of them

```bash
$ find . \( -name ".git*" -o -name .Rproj.user \) -prune -o -type f -print0 | xargs -0 dos2unix -ic
./inst/CITATION
./inst/testdata/dat20101314.csv
./misc/issues/glmernb_foraging_subset.csv
./misc/lme4_GH691_Dataone.csv
```
I left them alone.

**Still more changes to come.**

@bbolker The latest changes make the version 1.1-35.6 for consistency with current master.  That's probably a mistake; I'll make it 1.1-36 before I'm done.  But, as I said, I'm not sure R's version dependency machinery will interpret `>=1.1-36` as intended.  I'm not sure why you're hesitant to use semantic versioning: they aren't running out of numbers, and you can be like the browser makers by incrementing even the major version number with abandon :)

RossBoylan:
> * Running now, I don't get any errors at `LME4_TEST_LEVEL=10` (AFAIK all of the test-level statements are written as inequalities, so this should encompass level 2). Can you double check?

I got quite a few errors on my branch with `RStudio`, and so to isolate things I worked with the current master from the official github repo #bfd7a44d0a718fff090412871504858559a0829f.  Working in `Powershell` I set `LME4_TEST_LEVEL=2`, added the directory with the `R` binaries to `PATH`, built a tarball and finally checked, getting an error in `simulate.R`.

I did `LME4_TEST_LEVEL=2` because when I set it to 10 in `RStudio` the check took 22 minutes to complete, failing, like this one, in `simulate.R`.  Would you expect it to be that slow?

```ps1
C:\Users\rdboylan\Documents>Rcmd check lme4_1.1-35.6.tar.gz
* using log directory 'C:/Users/rdboylan/Documents/lme4.Rcheck'
* using R version 4.4.0 (2024-04-24 ucrt)
* using platform: x86_64-w64-mingw32
* R was compiled by
    gcc.exe (GCC) 13.2.0
    GNU Fortran (GCC) 13.2.0
* running under: Windows Server 2019 x64 (build 17763)
* using session charset: ISO8859-1
* checking for file 'lme4/DESCRIPTION' ... OK
* this is package 'lme4' version '1.1-35.6'
* package encoding: UTF-8
* checking package namespace information ... OK
* checking package dependencies ... OK
* checking if this is a source package ... OK
* checking if there is a namespace ... OK
* checking for executable files ... OK
* checking for hidden files and directories ... OK
* checking for portable file names ... OK
* checking whether package 'lme4' can be installed ... OK
* used C++ compiler: 'G__~1.EXE (GCC) 13.2.0'
* checking installed package size ... NOTE
  installed size is  7.1Mb
  sub-directories of 1Mb or more:
    R          1.5Mb
    doc        2.2Mb
    libs       1.3Mb
    testdata   1.3Mb
## etc, all good until
* checking for unstated dependencies in 'tests' ... OK
* checking tests ...
  Running 'AAAtest-all.R'
  Running 'HSAURtrees.R'
## all good until
  Running 'respiratory.R'
  Running 'simulate.R'
 ERROR
Running the tests in 'tests/simulate.R' failed.
Last 13 lines of output:
  +                newparams=list(theta=getME(fm1,"theta"),
  +                beta=fixef(fm1),
  +                sigma=sigma(fm1)),
  +                family=gaussian,
  +                seed=101)[[1]]
  + stopifnot(all.equal(s7,s1))
  +
  + ## TO DO: wider range of tests, including offsets ...
  +
  + }# R >= 3.0.0
  + } ## testLevel>1
  boundary (singular) fit: see help('isSingular')
  Loading required namespace: mlmRev
  Error: `simulate(...)` did not produce any messages.
  Execution halted

>     * I'm not sure about the vignette - it seems like it might belong in `misc/notes` instead? (This is `.gitignore` but you can force it in with `git add -f`)

That sounds like a better spot; I've moved it there.  I'll push the changes as soon as I send this, but it remains a work in progress.

**Comment by RossBoylan** [comment]
I'd like to move this along.  From my perspective, here's where this stands:

1. @mmaechler requested fixing the line endings, which I have done.  Maybe there's some button in the github gui I'm supposed to press to signal I have done so?

2. I tried to run test code, but encountered errors that may or may not be related to my changes, and which I requested feedback on.

I would like to be able to run a clean test suite, even at higher levels of `LME4_TEST_LEVEL` cleanly before declaring victory.

Finally, the base code has moved on, and this branch needs to be updated to track it.

**Comment by bbolker** [comment]
A few things:

## conflicts 

I tried to fix the conflicts here, but I don't immediately see how to do it without doing the merge itself. 

* the reported conflicts in DESCRIPTION and inst/NEWS.Rd appear to be from your changing the version number from 1.1-36 to 1.1.36. Please do *not* do this: we have been using the x.y-z format up until now, and I don't see a reason to change. (From `?as.package_version`:

> Numeric versions are sequences of one or more non-negative
     integers, usually (e.g., in package ‘DESCRIPTION’ files)
     represented as character strings with the elements of the sequence
     concatenated and separated by **single ‘.’ or ‘-’ characters** [emphasis added] 

* ` tests/testthat/test-summary_testlevel_1.rda ` is unimportant: it's output generated by `tests/testthat/test-summary.R`, and not actually used for anything else ...

## test failures

I believe this is where the tests are failing (this isn't actually what's in `tests/simulate.R`, but a minimal abstracted version).

```r
library(lme4)
L <- load(system.file("testdata/lme-tst-fits.rda",
                      package="lme4", mustWork=TRUE))
gm1 <- fit_cbpp_1
form <- ~ (1|herd) + period
tt <- getME(gm1,"theta")
bb <- fixef(gm1)
try(simulate(form,newdata=model.frame(gm1),
             newparams=list(theta=setNames(tt,"abc"),
                            beta=fixef(gm1)),
             family=binomial,
             weights=rowSums(model.frame(gm1)[[1]]),
             seed=101))
try(simulate(form,newdata=model.frame(gm1),
               newparams=list(theta=tt,
               beta=setNames(bb,c("abc",names(bb)[-1]))),
               family=binomial,
               weights=rowSums(model.frame(gm1)[[1]])))
```

You should get the following errors from the last two lines (not actually errors in this case because I've `try`-wrapped them):

> Error in matchNames(theta, tnames(object), "theta") : 
  mismatch between 'theta' parameter vector names and internal names (herd.(Intercept))

> Error in matchNames(beta, colnames(getME(object, "X")), "beta") : 
  mismatch between 'beta' parameter vector names and internal names ((Intercept),period2,period3,period4)

I'm not sure why you're not. I can see why this might have slipped through the cracks/not gotten tested on many different platforms because it's not done at the default test level, but I can't see any reason this particular test should be platform-sensitive ...

And yes, test level 10 is expected to be very slow - it's where we put all the tests that take a long time to run!

RossBoylan:
> A few things:
> ## conflicts
> 
I can clean those up when the other issues are done.  You've had some releases since I made the patches, and so they are a bit out of date.  But the version number is a bit more substantive an issue.

> I tried to fix the conflicts here, but I don't immediately see how to push my changes back to your branch (the instructions that GitHub gives appear to merge the pull request with the resolved conflicts into the master branch, and I don't want to do that yet ...)
> 
>     * the reported conflicts in DESCRIPTION and inst/NEWS.Rd appear to be from your changing the version number from 1.1-36 to 1.1.36. Please do _not_ do this: we have been using the x.y-z format up until now, and I don't see a reason to change.  ...

1.1.36 was your [proposal](https://github.com/lme4/lme4/pull/804#issuecomment-2351786294), motivated by my concern about being able to specify a versioned dependency from another package to assure getting the new feature.  Plus you thought my initial use of 1.2 was too big a jump, I guess.

...


> 
> 
> ## test failures
> 
> I believe this is where the tests are failing (this isn't actually what's in `tests/simulate.R`, but a minimal abstracted version).
> 
> ```r
> library(lme4)
> L <- load(system.file("testdata/lme-tst-fits.rda",
>                       package="lme4", mustWork=TRUE))
> gm1 <- fit_cbpp_1
> form <- ~ (1|herd) + period
> tt <- getME(gm1,"theta")
> bb <- fixef(gm1)
> try(simulate(form,newdata=model.frame(gm1),
>              newparams=list(theta=setNames(tt,"abc"),
>                             beta=fixef(gm1)),
>              family=binomial,
>              weights=rowSums(model.frame(gm1)[[1]]),
>              seed=101))
> try(simulate(form,newdata=model.frame(gm1),
>                newparams=list(theta=tt,
>                beta=setNames(bb,c("abc",names(bb)[-1]))),
>                family=binomial,
>                weights=rowSums(model.frame(gm1)[[1]])))
> ```
> 
> You should get the following errors from the last two lines (not actually errors in this case because I've `try`-wrapped them):
> 
> > Error in matchNames(theta, tnames(object), "theta") :
> > mismatch between 'theta' parameter vector names and internal names (herd.(Intercept))
> 
> > Error in matchNames(beta, colnames(getME(object, "X")), "beta") :
> > mismatch between 'beta' parameter vector names and internal names ((Intercept),period2,period3,period4)
> 
> I'm not sure why you're not. I can see why this might have slipped through the cracks/not gotten tested on many different platforms because it's not done at the default test level, but I can't see any reason this particular test should be platform-sensitive ...

So the problem is that the code running in my check failed to fail in the expected place.

I'll dig into this more.  Maybe it's an MS-Windows thing, but, like you, I find it hard to imagine how it could be, or something more specific about my environment.

> 
> And yes, test level 10 is expected to be very slow - it's where we put all the tests that take a long time to run!

Glad to hear that.

bbolker:
A few follow-up comments:

* to be clear, the point I was making above was about the difference between "1.1.36" (what was in your code) and "1.1-36" (the desired version number, in our code). Now we should probably actually be bumping to 1.1-37 (still with a hyphen), to make your dependency management work
* with some work we could set things up to run with test level 2 on r-hub/github actions on a Windows platform, but it would be handy if you're willing to take an initial pass to see if anything obvious pops out
* if I were you I think I would probably re-sync my branch with the upstream `lme4` master (and sort out any merge conflicts) first, since you'll have to sort these out sooner or later ...

RossBoylan:
While trying to work on the original problem I encountered a new one: in some environments the package would not compile.  To anticipate my theories at the bottom of this message, my suspicion is that when `NDEBUG` is not defined the resulting code after preprocessing is missing the definition of at least one function.

# Replicate Original Problem with Latest Code

I got current master (2c6fdff) from github, set LME4_TEST_LEVEL=2, ran `Rcmd build`, and finally `Rcmd check` on the resulting tarball.  This was all in Powershell, outside of `RStudio`.  The check failed in the same way as before.

# Compilation Failure

Wanting to trace the execution of the offending test script, I opened `RStudio` on the project directory.  In its console I executed `load_all()`.  This is from `devtools`, which I load automatically.  It tried to build the package and failed.

Here's the build that succeed, from `00install.out` in the `.Rcheck` directory (`Powershell`):
```sh
g++ -std=gnu++17  -I"C:/PROGRA~1/R/R-44~1.0/include" -DNDEBUG -I. -DNDEBUG -DEIGEN_DONT_VECTORIZE -I'C:/Program Files/R/R-4.4.0/library/Rcpp/include' -I'C:/Program Files/R/R-4.4.0/library/RcppEigen/include' -I'C:/Program Files/R/R-4.4.0/library/Matrix/include'   -I"C:/Programs/rtools44/x86_64-w64-mingw32.static.posix/include"     -O2 -Wall  -mfpmath=sse -msse2 -mstackrealign  -c predModule.cpp -o predModule.o
```

And here's the one that failed (`RStudio`):
```sh
> load_all()
i Loading lme4
i Re-compiling lme4 (debug build)
-- R CMD INSTALL -------------------------------------------------------------------------------------
-  installing *source* package 'lme4' ...
   ** using staged installation
   ** libs
   using C++ compiler: 'G__~1.EXE (GCC) 13.2.0'

# several successful compilations omitted

g++ -std=gnu++17  -I"C:/PROGRA~1/R/R-44~1.0/include" -DNDEBUG -I. -DNDEBUG -DEIGEN_DONT_VECTORIZE -I'C:/Program Files/R/R-4.4.0/library/Rcpp/include' -I'C:/Program Files/R/R-4.4.0/library/RcppEigen/include' -I'C:/Program Files/R/R-4.4.0/library/Matrix/include'   -I"C:/Programs/rtools44/x86_64-w64-mingw32.static.posix/include"     -O2 -Wall -gdwarf-2 -mfpmath=sse -msse2 -mstackrealign  -UNDEBUG -Wall -pedantic -g -O0 -fdiagnostics-color=always -c predModule.cpp -o predModule.o

In file included from 1mC:/Programs/rtools44/x86_64-w64-mingw32.static.posix/lib/gcc/x86_64-w64-mingw32.static.posix/13.2.0/include/c++/cassert:44,
                    from 1mC:/Program Files/R/R-4.4.0/library/RcppEigen/include/Eigen/Core:84,
                    from 1mC:/Program Files/R/R-4.4.0/library/RcppEigen/include/Eigen/Dense:1,
                    from 1mC:/Program Files/R/R-4.4.0/library/RcppEigen/include/RcppEigenForward.h:28,
                    from 1mC:/Program Files/R/R-4.4.0/library/RcppEigen/include/RcppEigen.h:25,
                    from 1mpredModule.h:12,
                    from 1mpredModule.cpp:8:
   lme4CholmodDecomposition.h: In instantiation of '1mvoid lme4::lme4CholmodDecomposition<_MatrixType, _UpLo>::1;32msolveInPlace(const Eigen::MatrixBase<OtherDerived>&, int) const 5m[with OtherDerived = Eigen::Matrix<double, -1, -1>; _MatrixType = Eigen::SparseMatrix<double, 0, int>; int _UpLo = 1]':
   1mpredModule.cpp:127:37:   required from here
   1mlme4CholmodDecomposition.h:59:39: 1;31merror: '1mIndex' is not a member of '1mlme4::lme4CholmodDecomposition<Eigen::SparseMatrix<double, 0, int> >::Base' {aka '1mEigen::CholmodDecomposition<Eigen::SparseMatrix<double, 0, int>, 1>'}
      59 |             eigen_assert(1;31m(Base::Index)(factor()->n) == other.rows());
         |                          1;31m~~~~~~~~~~~~~^~~~~~~~~~~~~
   1mC:/Program Files/R/R-4.4.0/library/RcppEigen/include/Eigen/src/Core/util/Macros.h:1037:25: 1;36mnote: in expansion of macro '1meigen_plain_assert'
    1037 | #define eigen_assert(x) 1;36meigen_plain_assert(x)
         |                         1;36m^~~~~~~~~~~~~~~~~~
   1mlme4CholmodDecomposition.h:59:13: 1;36mnote: in expansion of macro '1meigen_assert'
      59 |             1;36meigen_assert((Base::Index)(factor()->n) == other.rows());
         |             1;36m^~~~~~~~~~~~
   lme4CholmodDecomposition.h: In instantiation of '1mvoid lme4::lme4CholmodDecomposition<_MatrixType, _UpLo>::1;32msolveInPlace(const Eigen::MatrixBase<OtherDerived>&, int) const 5m[with OtherDerived = Eigen::Map<Eigen::Matrix<double, -1, 1> >; _MatrixType = Eigen::SparseMatrix<double, 0, int>; int _UpLo = 1]':
   1mpredModule.cpp:191:25:   required from here
# very similar errors repeat several times
```

The failure adds `-gdwarf-2 -UNDEBUG -pedantic -g -O0  -fdiagnostics-color=always` to the compilation options (toward the end).  Note the failing build announces itself as a debug build at the start.  So the first part of the line has production options (`-02 -DNDEBUG`) before the later options enabling debugging (`-UNDEBUG -g -O0`).  I recall that later options win, but I haven't checked it.

# Theories

  1. The code defined when `NDEBUG` is off is missing the definition of `Index`.
  2. `-pedantic` is causing trouble, perhaps turning warnings into failures.
  3. Options that control compiling and linking, apparently downstream of the error reported, are somehow responsible (`-gdwarf-2 -g -O0`.
  4. `-fdiagnostics-color=always` is responsible for the junk appearing in the log, e.g., `1mIndex`.  I assume the first 2 characters are misguided attempts at control of the display.  But it's hard to imagine this is more than a cosmetic issue.
  5. Other less explicit differences, e.g., in environment variables, are missing things up.  My system has things like Intel Math libraries that might interfere,  my current installation of a modified `lme4` which has version 1.2, git for Windows (with its own MSYS environment) and cygwin, both of which might hijack some Unix commands. RTools44 is installed, definitely being used in both builds, and I think is expected.  Not to mention that `RStudio` is a bit of a world unto itself.

Unfortunately for my preferred theory (1) the symbol NDEBUG does not occur in `predModule.{h,cpp}` or `lme4CholmodDecomposition.h`.  It does occur in spots in the `RcppEigen` headers, but nothing with an obvious connection to the problem.  Then again, I didn't even find where `Index` was defined.

bbolker:
I'm sorry you're having so much trouble, but I'm not sure I'm going to be able to help very much: my "solution" would be, since (1) compilation *does* work with the standard `R CMD build` outside of RStudio/`devtools`, (2) it *does* work on all of the systems CRAN uses for testing, (3) none of our other users have reported similar compilation problems [although admittedly a small fraction of users are trying to compile from source on Windows], that this is an edge case I'm not too worried about.<sup>*</sup>. I have experienced problems with `devtools::load_all()` and compiled code in the past (not this particular problem, admittedly), presumably because some of the machinery that `devtools` uses to shim/mimic the full `R CMD build/R CMD INSTALL/library()` process ...

Does `R CMD INSTALL <package_dir>` work outside of RStudio? If it does, that will leave the compiled objects in the `src` directory so you should be able to `devtools::load_all()` without recompiling ... I'm also curious whether `Rscript -e "devtools::load_all('package_dir')"` works from a shell prompt, i.e. outside RStudio ... (i.e. perhaps there is an RStudio + devtools + compilation tools interaction?)

It might also make sense to ask on `r-sig-mixed-models@r-project.org`: maybe someone with a Windows system can replicate/help troubleshoot your problem ...

---
<sup>*</sup> I have to admit that CRAN's relentless pursuit of issues that arise on obscure platforms (SPARC Solaris anyone?) has contributed over the years to the robustness of the R ecosystem, but there's only so much I can do ...

bbolker:
FWIW I can confirm that (1) `devtools::load_all()` fails under RStudio if full compilation is necessary (i.e. there are no binary `.o`/`.so`/whatever files in the `src` directory); (2) first running `R CMD INSTALL` from the command line within the directory, then running `devtools::load_all()` from within RStudio, works fine (because no recompilation is necessary).

RossBoylan:
I think I've isolated the problem.  Removing `-UNDEF` from the `g++` command that had the error, was sufficient to stop the error.  So then the question is why.

The source of the failure seems to be line 59 of `lme4CholmodDecomposition.h`:
```cpp
            eigen_assert((Base::Index)(factor()->n) == other.rows());
```
because `Base` has no member `Index`.   The `eigen_assert` only produces code if it is in debug mode (in the sense that `NDEBUG` is not a defined macro), and so it is only alive to cause a compiler error then.  I verified this by examining the preprocessor output (`g++ -E` instead of `-o`).

So either the reference to `Base::Index` is simply in error, or it really should exist, but something else is blocking the creation of it.

`eigen` uses `Index` extensively as a type alias, either in `C++ template<>` parameters or in `C`-style `typedef`s.  There is an `Eigen::Index`, at least in some contexts, e.g., `DenseStorage.h`.  From `EigenBase.h`
```cpp
namespace Eigen {
## ...
template<typename Derived> struct EigenBase
{
## ...
  /** \brief The interface type of indices
    * \details To change this, \c \#define the preprocessor symbol \c EIGEN_DEFAULT_DENSE_INDEX_TYPE.
    * \sa StorageIndex, \ref TopicPreprocessorDirectives.
    * DEPRECATED: Since Eigen 3.3, its usage is deprecated. Use Eigen::Index instead.
    * Deprecation is not marked with a doxygen comment because there are too many existing usages to add the deprecation attribute.
    */
  typedef Eigen::Index Index;
```
But template `typenames`, I believe, are not available to consumers of the template.

`Meta.h` seems to define `Eigen::Index` which eventually leads to `ptrdiff_t`.  Since there are almost 11,000 matches to the case-sensitive regex `\bIndex\b` in the `RcppEigen` source it's hard to be sure!

My suspicion is that `Base::Index`  really is not defined where the error occurs; `Eigen::Index` seems more likely to work, but as it is not sensitive to the context in which it is used it might not be appropriate either.

As a reminder, this is all a detour on the road to isolating the original problem with the test code, which used a non-debug build.

bbolker:
Thanks. Do you have any suggestions about how to move forward with this? To be honest my preference would be to open an issue for reference, but not worry about it/invest further debugging time otherwise. 

For what it's worth `git blame` attributes these lines to ["patches from K Millar for unmodified Eigen"](https://github.com/lme4/lme4/commit/dcaa5f5879e77d644865106219b8cee6900d6cf9) from 2014. The only place "Millar" appears in the issues list is [here](https://github.com/lme4/lme4/issues/209); there is a long thread in the `lme4-authors` mailing list from May 2014 about issues with RcppEigen that were causing memory-test issues on CRAN. Karl Millar (at Google at the time) donated the patch. I don't know whether it's worth (1) asking him about this line; (2) asking on the Rcpp mailing list; (3) just commenting this line out ... ???

RossBoylan:
I’m happy with tagging it as a separate issue and commenting out the line, which is what I did.
I assume the test was there for a reason, and so it might be good to chase that down… someday.

With that one change I’m now able to build and check with LME4_TEST_LEVEL=2 outside of RStudio, load_all() works, and the original problem with failing tests seems to have gone away.  That last one seems weird, since if my theory is correct the commented out line wasn’t generating any code in the regular build.  So commenting it out shouldn’t matter, and I should still have the original problem.  I need to dig into that more.

Now that I can load_all() I was able to execute pieces of the simulate.R test file, in particular the code inside the expect_error() functions.  Both worked as expected, producing an error with the expected string.

Maybe I’m doing something different, or maybe my thought `eigen_assert` produces no code outside debug mode is wrong; maybe it produces some code, but doesn’t do the inner assert.  But then it would further need to be the case that the code that was produced by eigen_assert was responsible for the original bug (the non-occurrence of an expected error message during testing), which seems unlikely.

If you want it, here was my commented commenting out:
            /* The code at the end of this comment produces
             * error: ‘Index’ is not a member of ‘lme4::lme4CholmodDecomposition<Eige...
             * It only produces the error in debug mode (-UNDEBUG); otherwise it
             * produces no code at all, and thus no error.
             * devtools::load_all() is one of the ways to get debug code.
             * So I'm removing the code from the active source.
             * Ross Boylan
             *
              eigen_assert((Base::Index)(factor()->n) == other.rows());
              */

bbolker:
OK. If the peripheral issues are now gone and/or deemed ignorable, please bring your pull request up to date so that it's minimal and non-conflicting with the master branch ...

bbolker:
Any updates on this?

RossBoylan:
I think it's ready to go with the changes I just pushed.

Merged with latest `master`.  Added my changes as v1.1-38, which you may wish to make part of 1.1-37, which is already there.  I remain concerned that the machinery for dependency management will have trouble dealing with stuff after the `-` in the version, but `utils::compareVersion()` seemed able to cope, and so I went with it.

builds and checks clean (on MS-Windows) with LME4_TEST_LEVEL=10.  I believe this is partly because 260cdebfdd41929422d6a1dda78554ab3a315965 removed the tests for simulate that were producing the 
```
Error: `simulate(...)` did not produce any messages.  
```
failure.  I got that in 175615d91e9ccf833ef463bc80b430fea0e2f69e, just before merging the latest.  After (e9a8cad402c50e119a4e43c7155b544a79c0df47) the merge things were good.

The revisions include tests for the new feature; like most of the `simulate` tests they only run if `LME4_TEST_LEVEL>1`.

I reverted `test-summary_testlevel_1.rda` to upstream.

My scan for `DOS` line endings shows nothing I touched, and so I think there shouldn't be a lot of noise differences:
```sh
$ find . \( -name ".git*" -o -name .Rproj.user \) -prune -o -type f -print0 | xargs -0 dos2unix -ic
./.vscode/settings.json
./inst/CITATION
./inst/testdata/dat20101314.csv
./lme4.Rproj
./misc/issues/glmernb_foraging_subset.csv
./misc/lme4_GH691_Dataone.csv
```

Issue #481, opened 2018-08-24
weights (and/or offset) variables overwritten in global env


marjoleinF:
First of all: many thanks for creating lme4! 

I noticed, when weights and offset arguments are supplied to lmer(), weights and offset variables are created in the global environment. This problem is often harmless, as noted in [#191](https://github.com/lme4/lme4/issues/191), but seems exacerbated when a weights variable exists in the global environment and we apply a function which calls lmer() with a weights argument. Then the weights variable in the global environment gets overwritten. An example is shown [here](https://github.com/marjoleinF/misc/blob/master/weights_and_offset_bug_report_lme4.md#lmer-creates-weights-and-offset-variables-in-global-environment).

bbolker:
This is a good point ... will try to dig in and take a look sometime soon, although pull requests fixing the problem are certainly welcome!  Thanks.

bbolker:
proximal problem is [here](https://github.com/lme4/lme4/blob/master/R/modular.R#L365-L367) (and [here](https://github.com/lme4/lme4/blob/master/R/modular.R#L651-L661) (pasted below for convenience). If we can come up with an incantation for evaluating the model frame that correctly allows it to see the values that have been passed (separately) as `weights` and `offset`, without making any copies ... is there some kind of "enclosing environment" magic we can use?

```
 fr.form <- subbars(formula) # substitute "|" by "+"
    environment(fr.form) <- environment(formula)
    ## model.frame.default looks for these objects in the environment
    ## of the *formula* (see 'extras', which is anything passed in ...),
    ## so they have to be put there ...
    for (i in c("weights", "offset")) {
        if (!eval(bquote(missing(x=.(i)))))
            assign(i, get(i, parent.frame()), environment(fr.form))
    }
    mf$formula <- fr.form
fr <- eval(mf, parent.frame())
```

dmbates:
Offhand I don't know of a way of creating an "enclosing environment".  One of the reasons for changing from a multiple formula model specification in `nlme` to a single formula incorporating both fixed-effects and random-effects terms in `lme4` was to have a single formula environment.  Of course, the `weights` and `offset` were always the wild cards in this scheme.

bbolker:
this should be solved by the PRs mentioned above.

Issue #191, opened 2014-03-13
RFC: offset/weight handling (mixlm compatibility problem)


bbolker:
In the current version (1.1-5) we have managed to break compatibility with the `mixlm` package.

**tl;dr** I've fixed the problem, sort of, but this exposes a bit of the code that is fragile; there might be a much less ugly way to solve the original problem ... ?

```
library(mixlm)
packageVersion("mixlm") ##  1.0.0
packageVersion("lme4")  ##  1.1.5
set.seed(101)
dataset   <- data.frame(y=rnorm(8),
       x=factor(c(rep(1,4),rep(0,4))),
       z=factor(rep(c(1,0),4)))
GLM       <- glm(y  ~ x+r(z), family=gaussian(identity), data=dataset)
## Error in get(i, parent.frame()) : 
##   argument "weights" is missing, with no default
## In addition: Warning message:
## In lme4::glmer(rw$reml.formula, data, family = family, REML = REML,  :
##   calling glmer() with family=gaussian (identity link) as a shortcut to lmer() is deprecated; please call lmer() directly
```

Proximally, the problem occurs (actually "occurred -- now fixed, see below) in `lFormula` (in `R/modular.R`), at l. 295:

```
## model.frame.default looks for these objects in the environment
## of the *formula* (see 'extras', which is anything passed in ...),
## so they have to be put there ...
for (i in c("weights", "offset")) {
     assign(i,get(i,parent.frame()),environment(fr.form))
}
```

No-one has reported problems with this in the past, but any time I muck around with getting things out of parent frames I accept that it can be fragile.  The particular problem here is that normally when `weights` is missing we end(ed) up doing something silly but harmless -- the `weights()` function from `stats` was copied.  In this case we are in trouble because the parent frame is a function with a `weights` argument that might be missing (no default/not specified by user), so trying to evaluate it throws an error.

This code is now fixed as follows:

```
for (i in c("weights", "offset")) {
    if (!eval(bquote(missing(x=.(i)))))
        assign(i,get(i,parent.frame()),environment(fr.form))
}
```

so that it doesn't try to assign the variable if it hasn't been passed to `lFormula`.

However, I wonder if there is an entirely better way to do this.  Indeed, I don't remember why it's written this way in the first place, but there clearly must have been a reason why we had to use `get(i,parent.frame())` (rather than just `get(i)`) ...

PR #811, opened 2024-10-15
Fix lookup of optional arguments in lFormula()

klmr:
`lFormula()` was looking up the argument names `weights` and `offset` in the *caller scope* instead of the current scope. This will fail unless the caller happens to have defined local variables of those exact names.

However, it usually fails *silently*, since `get()` by default uses `inherits = TRUE` and searches through the chain of parent environments, which usually include attached packages. `lFormula()` therefore finds `stats::weights` and `stats::offset`, and by pure (bad) luck these don’t trigger an error even though they’re not of the expected type (functions instead of numeric vectors).

See klmr/box#373.

(I’m aware that this code also contains another issue, discussed in #481, but fixing that is more involved: it is not fixed by this PR.)

### DIFF
--- a/R/modular.R
+++ b/R/modular.R
@@ -374,7 +374,7 @@ lFormula <- function(formula, data=NULL, REML = TRUE,
     ## so they have to be put there:
     for (i in c("weights", "offset")) {
         if (!eval(bquote(missing(x=.(i)))))
-            assign(i,get(i,parent.frame()),environment(fr.form))
+            assign(i,get(i),environment(fr.form))
     }
     mf$formula <- fr.form
     fr <- eval(mf, parent.frame())


bbolker:
Thank you very much!  Any chance you can provide a test for this as well? (If not I'll accept it anyway.)

I need to be careful about this: there is often a reason that things are done in this weird way. This assignment was introduced [here](d208e9a3593727b55ef007ec6); unfortunately I don't have a test case explaining why I chose to do it this way; i.e., I don't know whether I chose to evaluate in the parent frame because I had a particular case where that was required, or because I was guessing by analogy with other machinery that uses this.

These scoping issues are a huge pain - there are many different contexts in which people can end up calling these functions, and after many years I still don't have an adequate mental model of the right way to handle the lookups (or whether there is even a single right way).

klmr:
@bbolker That link is dead for me.

I’ll see if I can create a simple test case.

bbolker:
Maybe [here](https://github.com/lme4/lme4/commit/d208e9a3593727b55ef007ec6dac829e17e1216e) instead (admittedly there's nothing there that explains *why* I implemented it that way ...)  Will try to come up with some other realistic test cases (e.g. define formula in one scope, weights in another, pass them through to a third function to do the fit ...)

klmr:
Actually after playing a bit with the code it appears that [the comment](https://github.com/lme4/lme4/commit/d208e9a3593727b55ef007ec6dac829e17e1216e#diff-6102084076fc4e633832435611811438b40b4da2ee6205b101aab0047ee26761R499-R501) misstates how `model.frame` works — and consequently the entire subsequent code is unnecessary. It can be removed wholesale, and this will also solve #481.

It’s true (as per the documentation — and also the implementation!) that `model.frame.default` looks up the values of `weights` and `offset` in the environment of the formula. But it doesn’t look for names called `weights`/`offset`. Instead, it (correctly) looks for their assigned expressions. So, in the case of the linked MRE (`lme4::lFormula(form, data = tbl, weights = tbl$hp)`), it evaluates `tbl$hp` in the scope of the formula. And this will be successful — it completely ignores the names assigned inside `lFormula` (unless we had a different variable of that name in the environment where `formula` is defined, which will be overwritten).

However, this means that if the formula is defined in a *different scope*, the situation changes:

```r
my_weights = rep(1, 6L)
formula = mpg ~ disp + (1 | cyl)

local({
  my_weights = rep(2, 6L)
  lme4::lFormula(formula, data = head(mtcars), weights = my_weights)$fr
})
```
⇒
```
                   mpg disp cyl (weights)
Mazda RX4         21.0  160   6         1
Mazda RX4 Wag     21.0  160   6         1
Datsun 710        22.8  108   4         1
Hornet 4 Drive    21.4  258   6         1
Hornet Sportabout 18.7  360   8         1
Valiant           18.1  225   6         1
```

— note that this is using `my_weights` from the environment in which `formula` was defined, not the one from the calling scope. From the comment it seems that this is the situation that `lFormula` wants (but fails!) to avoid, but it’s unclear whether that’s actually what should be done: after all, it’s *intentional* that R formulas carry their environment. So maybe that’s what the user wanted?

And the following also fails (currently, but it would still fail if the assignment is removed):

```r
formula = mpg ~ disp + (1 | cyl)

local({
  tbl = head(mtcars)
  lme4::lFormula(formula, data = tbl, weights = tbl$hp)$fr
})
```

(Note that this can be fixed by passing `weights = hp` instead of `weights = tbl$hp`; however, the documentation gives no indication that this is supported or intended: it doesn’t mention NSE in the context of the `weights` or `offset` arguments.)

So in my understanding there are two viable options (and one I’d consider a non-starter) here:

1. **Remove the assignment completely.** This is a straightforward fix, and it restores the default scoping behaviour of R formulas as designed. However, in the two situations shown above this *might* lead to surprising behaviour for users, if they expect the arguments of `lFormula()` to refer to objects in the current scope, rather than being subject to non-standard evaluation.
2. **Reassign the formula environment.** That is, assign `parent.frame()` as the environment of the local copy of `formula`, and document the fact that the environment of the `formula` argument is going to be ignored (because it is reassigned internally). This breaks the way formulas normally work, but is probably the behaviour most users expect.
3. (**Fiddle with the formula environment.**) We could also attempt a hybrid where the formula environment is kinda-preserved but the calling scope is added as a parent to enable lookup in *both*. IMO this gets *very* messy (especially if the formula is defined in the global environment, because then we’d need to *attach* the calling frame!).

Either way, the documentation should be updated across all relevant functions to clarify how these arguments are evaluated.

What do you think?

bbolker:
I think I'm nervous about this and would like more eyes on it. I'm not saying the previous code was correct,  or that the new code isn't better, but the code that's there was introduced for a reason (even if that reason is extremely poorly documented!) and has been being used in **many** contexts for a long time.

See for example [here](https://github.com/lme4/lme4/issues/191)

To do the archaeology properly I might have to go back into the pre-GitHub SVN repo to see where this code originated ...

@mmaechler, thoughts?

annahuynhly:
Tried to instead fix using: https://github.com/lme4/lme4/pull/961

PR #814, opened 2024-11-19
The Nelder_Mead function accepts the trace option. Also, fn can return NA to stop

pbosetti:
The trace option in the `Nelder_Mead` function allows to get a data frame with the intermediate points. Also, if/when the function to be optimized returns a NA, the iteration stops cleanly.

The latter change allows to use the algorithm interactively, which may prove useful if the function to be optimized is actually an external simulation that you cannot control programmatically. In such a circumstance, you may call:

```r
Nelder_Mead2(\(x) {
    cat("Value at:", x, "\n")
    readline() %>% as.numeric()
    }, p0, trace=T)
```

and just enter `NA` to stop the process earlier, while still keeping trace of the evolution.

Admittedly, it would be great to allow to start the optimization from a simplex, instead from a single point, for this would allow to restart an optimization that has been stopped earlier. But I need to study the codebase more deeply to figure out if that can be done.

### DIFF
--- a/R/optimizer.R
+++ b/R/optimizer.R
@@ -1,7 +1,11 @@
 ## --> ../man/NelderMead.Rd
 Nelder_Mead <- function(fn, par, lower=rep.int(-Inf, n),
-                        upper=rep.int(Inf, n), control=list()) {
+                        upper=rep.int(Inf, n), control=list(), trace=FALSE) {
     n <- length(par)
+    if (trace) {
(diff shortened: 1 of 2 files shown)


pbosetti:
Can do that. Another option would be to switch to "stop on FALSE", so that an NA can still be managed, but if the function explicitly returns a logical false the iteration stops. This way we avoid adding another option.
What do you prefer?

Issue #555, opened 2020-01-17
Possible problem with safe_chol?


pitakakariki:
I noticed that `safe_chol` can return non-triangular matrices in the semidefinite case. I don't think this is used in any exported code though?

```
V2 <- matrix(c(1600, 1160, 1760, 1160, 3442, 2806, 1760, 2806, 2836), 3)

V2
#      [,1] [,2] [,3]
# [1,] 1600 1160 1760
# [2,] 1160 3442 2806
# [3,] 1760 2806 2836

safe_chol(V2)
#          [,1]     [,2]    [,3]
# [1,] 19.77209 58.66856 47.8280
# [2,] 34.77160  0.00000 23.4197
# [3,]  0.00000  0.00000  0.0000

all.equal(crossprod(safe_chol(V2)), V2)
# [1] TRUE
```

I had a similar bug in my own code, I'm using this instead now (although I don't know if it's particularly numerically stable):

```
semi_chol <- function(V) {

    # V = U D U'
    E <- eigen(V)

    # V = B' B where B = sqrt(D) U'
    B <- diag(sqrt(pmax(E$values, 0))) %*% t(E$vectors)

    # V = R' Q' Q R, where Q R = B
    R <- qr.R(qr(B))

    # V = R' Q' Q R = R' R
    R
}

semi_chol(V2)
#      [,1] [,2] [,3]
# [1,]   40   29   44
# [2,]    0   51   30
# [3,]    0    0    0
```

r499y:
Hi @pitakakariki, I'm working on this issue. I see that for PSD matrices, the current implementation of safe_chol() returns a factor L such that crossprod(L) equals the original matrix, but L isn't strictly triangular. According to [Wikipedia](https://en.wikipedia.org/wiki/Cholesky_decomposition), a triangular factor for PSD matrices may not be unique or even enforced when the matrix is rank-deficient.

Would you prefer that we modify safe_chol() to always return a strictly triangular factor, or is it acceptable to document that this behavior is expected for PSD cases? I want to make sure my fix aligns with the project's goals.

pitakakariki:
I think I'll wait for Ben to weigh in on this!

It looks like the  function and the "## FIXME:" comment predate lme4's move to github, and now that the important Cholesky work is done on sparse matrices using CHOLMOD it might just be there as a "behind the scenes" convenience?

I do like the use of LDLT and an R implementation of that is probably better than adding another dependency - I've only ever used safe_chol on small examples.

PR #823, opened 2025-03-24
Fix safe_chol() for PSD matrices using LDL^T fallback (#555)

r499y:
This pull request addresses #555 regarding the behavior of `safe_chol()` when handling positive semidefinite (PSD) matrices. In the current implementation, when a PSD matrix is encountered, the fallback using a pivoted Cholesky decomposition sometimes returns a factor that is not strictly triangular, even though it satisfies `crossprod(L) == V`. This can lead to unexpected behavior in downstream code that assumes a standard (upper-triangular) Cholesky factor.

  
The function now first attempts a standard Cholesky decomposition. If that fails (indicating a PSD matrix), it falls back on an LDL^T decomposition. In this fallback, we compute the LDL^T factors (with appropriate tolerance checks for near-zero pivots) and then form an upper-triangular factor `U` as follows:
  
  ```
  U = diag(sqrt(D)) %*% t(L)
  ```
  
where `L` is the unit lower-triangular factor from the LDL^T decomposition and `D` is the corresponding diagonal vector. This ensures that:
  
  - `crossprod(U)` approximates the input matrix `V` within a specified tolerance, and  
  - `U` is strictly upper-triangular.

Additional tests have been added in `tests/testthat/test-vcconv.R` to ensure that:
  - For PSD matrices, `safe_chol()` returns a factor whose lower-triangular entries are all zero, and  
  - The reconstruction via `crossprod(L)` matches the original matrix within the defined tolerance.

  
There is nuance in the literature regarding the Cholesky decomposition of PSD matrices. According to [[Wikipedia](https://en.wikipedia.org/wiki/Cholesky_decomposition)](https://en.wikipedia.org/wiki/Cholesky_decomposition), while a unique triangular factor is guaranteed for positive definite matrices, the factorization in the semidefinite case is not unique. In theory, this means that a nontriangular factor might be acceptable. However, within the context of `lme4`, downstream functions and tests expect an upper-triangular factor.
 
- This fix is implemented following discussions with our GSOC mentor @bbolker.  
- Thanks to @pitakakariki for opening the issue and providing valuable input.

Please review the changes and let me know if further modifications are needed. Thank you!

---

### DIFF
--- a/R/vcconv.R
+++ b/R/vcconv.R
@@ -140,23 +140,63 @@ cov2sdcor  <- function(V) {
 ##     m
 ## }
 
-## attempt to compute Cholesky, allow for positive semi-definite cases
-##  (hackish)
-safe_chol <- function(m) {
-    if (any(is.na(m)) || all(m==0)) return(m)
-    if (nrow(m)==1) return(sqrt(m))
-    if (.isDiagonal.sq.matrix(m)) return(diag(sqrt(diag(m))))
-
-    ## attempt regular Chol. decomp
-    if (!is.null(cc <- tryCatch(chol(m), error=function(e) NULL)))
-        return(cc)
-    ## ... pivot if necessary ...
-    cc <- suppressWarnings(chol(m,pivot=TRUE))
-    oo <- order(attr(cc,"pivot"))
-    cc[,oo]
-    ## FIXME: pivot is here to deal with semidefinite cases,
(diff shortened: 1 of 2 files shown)


annahuynhly:
There is a weird error when running `devtools::document()` with this branch; seems to be related to files in `lme4CholmodDecomposition.h`...

Issue #227, opened 2014-08-08
Add option to automatically z-standardize predictors and dependent variable


mmeierer:
From my experience z-standardization - i.e. rescaling the variables - speeds up optimization and might help in case of convergence warnings. lmer already displays a message that rescaling might be necessary if the variables are on very different scales. 

In case of larger models it is cumbersome to do the z-standardization for every variable. **Having an option in the lmer command to automatically z-standardize all predictors all well as the dependent variable would be great.** Thus, the estimated parameters would be the standardized coefficients (which everybody is familiar with from linear regression).

If this option is chosen, it would be great if also the unstandardized coefficients are calculated (in the post estimation stage) as outlined by @bbolker on SO
http://stackoverflow.com/questions/23642111/how-to-unscale-the-coefficients-from-an-lmer-model-fitted-with-a-scaled-respon

[Feature Request]

rubenarslan:
+1 for this idea, I've wished for this often, because I have often have predictors on a naturally intuitive scale (e.g. years, siblings), so I don't want to scale and get standardised coefficients, but I also reaaally want my models to converge faster.

I also have the feeling that this warning happens very quickly in large datasets, at very small differences in scale. I get "Model is nearly unidentifiable: very large eigenvalue" already with data like below. Scaling the `nr.siblings` variable makes it go away, but the SD is only ~6 times that of the smallest continuous variable (dummy variables are marked *). 
This seems like a difference you will often encounter (and which is not usually the result of some scaling blunder). The warning is spot on though: With the predictor scaled the models takes 245 seconds, with the predictor just centered it takes 545 seconds!
The warning also disappears if I cut down the dataset from 85805 to 2000. This is probably obvious to you, but I was at first puzzled why I get more "problems" with more data.

```
                     mean   sd
children             4.44 5.40
center(spouses)      0.01 0.70
center(nr.siblings)  0.27 3.67
paternalage.mean     0.16 0.64
paternalage.diff    -0.09 0.56
male*                1.51 0.50
urban*               1.22 0.42
maternalage.factor*  1.53 0.85
paternalloss*        0.06 0.23
maternalloss*        0.06 0.25
birth.cohort*        3.03 1.38
```

PR #824, opened 2025-03-25
Add auto z_scale option #227

YancyLan:
**Issue Addressed:**
This change addresses [GitHub issue #227](https://github.com/lme4/lme4/issues/227), which suggested adding an option to automatically z-standardize predictors.
**How It Works:**
By including "zscale" in the check.scaleX vector, users can now set check.scaleX = "zscale" in their call to lmerControl (or the equivalent functions). When this option is selected (assuming the corresponding branch in the checkScaleX function is implemented), the model matrix's numeric columns are transformed using a z-standardization (subtracting the mean and dividing by the standard deviation).
**Note:**
This solution does not handle back-transformation since the issue specifies that back-transformation is not a concern.

### DIFF
--- a/R/lmerControl.R
+++ b/R/lmerControl.R
@@ -84,7 +84,7 @@ merControl <-
                              "silent.drop.cols", "warn+drop.cols",
                              "stop.deficient", "ignore"),
              check.scaleX = c("warning","stop","silent.rescale",
-                              "message+rescale","warn+rescale","ignore"),
+                              "message+rescale","warn+rescale","ignore","zscale"),
              check.formula.LHS = "stop",
              ## convergence options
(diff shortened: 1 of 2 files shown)


YancyLan:
**Issue Addressed:**
This change addresses [GitHub issue #227](https://github.com/lme4/lme4/issues/227), which suggested adding an option to automatically z-standardize predictors.
**How It Works:**
By including "zscale" in the check.scaleX vector, users can now set check.scaleX = "zscale" in their call to lmerControl (or the equivalent functions). When this option is selected (assuming the corresponding branch in the checkScaleX function is implemented), the model matrix's numeric columns are transformed using a z-standardization (subtracting the mean and dividing by the standard deviation).
**Note:**
This solution does not handle back-transformation since the issue specifies that back-transformation is not a concern.

Issue #770, opened 2024-02-18
make `isSingular` an S3 method


bbolker:
could add a default method based on https://github.com/easystats/performance/issues/681 (check minimum eigenvaues of covariance matrices) and move the current `isSingular` to `isSingular.merMod` ...

mmaechler:
Why? IIUC, the case there showed that *their* `check_singularity()` was clearly insufficient .. but not a deficiency in our `isSingularity()`, right?  Using a generic and methods is quite a bit of (low level computational) overhead.

bbolker:
The motivation for this is that I wanted to migrate `performance`'s `check_singularity` to a default method of `isSingular` so that it could more easily be used in `glmmTMB`.  Basically, I'd rather that tests for singularity get handled by individual packages rather than written by the `performance` authors, because authors of mixed model packages know more about singularity and how to test for it than the authors of `performance` do ...

I'm not particularly concerned about the overhead of testing for singularity, as it usually gets done after models are fitted (which would take way more time than the test itself).

I may be doing it wrong below (maybe the testing takes more overhead when the method is exported from a package?), but it doesn't seem to add much overhead in this example ...

```
library(lme4)

is_sing <- function(x) {
    UseMethod('is_sing')
}

is_sing.default <- function(x) {
    lme4::isSingular(x)
}

m1 <- lmer(Reaction ~ 1 + (Days|Subject), sleepstudy)

library(microbenchmark)
microbenchmark(isSingular(m1), is_sing(m1))
```

GregForkutza:
Hi all,

Just a quick note to say that I'm currently working on this issue as my submission for the Hard test for [this GSOC project proposal](https://github.com/rstats-gsoc/gsoc2025/wiki/Structured-covariance-matrices-for-lme4) including an eigenvalue-based method for isSingular() along with accompanying tests. I understand others are free to explore it as well, but I wanted to mark my interest in case it helps coordinate efforts or avoid duplicate work.

Thanks!

bbolker:
This is getting revived because of the preference for flexibility in the way we test singularity for new-style covariance structures. Checking whether any bounds (lower *or* upper) are active *may* be sufficient, but I think it will be worth it to specify covariance-type-specific methods (and then `isSingular.merMod` will be `any(vapply(getReCovs(object), isSingular, FUN.VALUE=logical(1)))`)

PR #825, opened 2025-03-25
Refactor isSingular into an S3 Generic #770

YancyLan:
**Issue Addressed:** This pull request refactors the isSingular function to be an S3 generic, allowing for method dispatch based on the class of the model object. Users and developers can now define class-specific behaviors for singularity checks.This method is to address issue [lme4 Issue #770](https://github.com/lme4/lme4/issues/770) #770 

**How it works:**
Existing code using isSingular() will continue to work as it now dispatches to isSingular.default if no class-specific method is available.

**Next Step:** Consider extending class-specific methods for additional model classes as needed.

### DIFF
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -263,7 +263,7 @@ export(vcov.merMod)
 S3method(vcov,summary.merMod)
 S3method(weights,merMod)
 S3method(xyplot,thpr)
-
+S3method(isSingular, default)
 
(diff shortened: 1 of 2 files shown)


annahuynhly:
Closing this pull request as the issue will be completed by this pull request instead: https://github.com/lme4/lme4/pull/826

PR #826, opened 2025-03-25
Refactor `isSingular()` into S3 generic with eigenvalue-based method

GregForkutza:
### Summary

This PR refactors `isSingular()` into a proper S3 generic and introduces a new method `isSingular.merMod()` for linear mixed models. The new method supports both the original Cholesky-based check and a new eigenvalue-based check using the `method` argument:

```r
isSingular(model, method = c("cholesky", "eigen"))
```

### Features
- Refactors isSingular() as an S3 generic.

- Moves original logic to isSingular.merMod(method = "cholesky").

- Adds an alternative method = "eigen" for rank-deficiency detection via eigenvalues.

- Updates lmerControl() to avoid relying on formals(isSingular)$tol.

- Adds unit tests for both singular and non-singular cases.

- Updates NEWS.Rd under Development version.

### Motivation

The Cholesky-based method can miss cases where the random-effects covariance matrix is nearly singular but all diagonal elements of the Cholesky factor are still above the threshold.

The new eigenvalue-based approach improves robustness by explicitly checking the minimum eigenvalue of the random-effects covariance matrix.

### Example Usage

```r
# Fit a mixed model
m <- lmer(Reaction ~ Days + (Days | Subject), data = sleepstudy)

# Check for singularity using eigenvalues
isSingular(m, method = "eigen")

# Or fall back to original behavior
isSingular(m, method = "cholesky")
```


R/isSingular.R
```diff
@@ -0,0 +1,30 @@
...
+  UseMethod("isSingular")
+}
+
+#' @export
+isSingular.merMod <- function(object, tol = 1e-6, method = c("eigen", "cholesky"), ...) {
+  method <- match.arg(method)
+  
+  if (method == "cholesky") {
+    cc <- getME(object, "ST")
+    diagvals <- lapply(cc, function(x) diag(x))
```
stevencarlislewalker, line 15: Small style point -- this line could be simpler as:
```
diagvals <- lapply(cc, diag)
```

GregForkutza: Good point — updated to use lapply(cc, diag) directly.

stevencarlislewalker: Thanks @GregForkutza .  Certainly would have been totally fine as it was, but a little easier to read this way.

annahuynhly:
@bbolker Should be safe to look at now!

Issue #468, opened 2018-04-25
Diagnostic plots


maciejdzialo:
Is there a chance to implement diagnostic plots similar to what we have with plot.lm() function? Right now only pearson residuals vs fitted plot is available, but it will be interesting to do diagnostics with additional scale-location plot, qq-plot, and residuals vs leverage plot. Thanks in advance!

bbolker:
Oh, now I see that I need to implement  `rstandard` and `cooks.distance` methods as well ...

bbolker:
This should be working now.  Test?

bbolker:
Maybe worth adding examples/cross-links in documentation? This stuff is definitely in the JSS paper/vignette (`vignette("lmer", package = "lme4")`), and in `?plot.merMod`, but I don't know how easy it is to   find if you don't know it's there.

Also partly superseded by `performance::check_model()` and `DHARMa` - but these be cross-referenced in the `?plot.merMod` page!

PR #829, opened 2025-05-11
Attempt at improving the documentation for plot.merMod.Rd.

annahuynhly:
My attempt at dealing with this current lme4 issue as stated here: https://github.com/lme4/lme4/issues/468 

In summary,
- very slightly modified lmer.Rd to mention diagnostic plots and refer to plot.merMod.Rd
- modified plot.merMod.Rd by adding references to \seealso, as well as the description header telling the user to visit \seealso. Also added links for external packages

### DIFF
--- a/man/lmer.Rd
+++ b/man/lmer.Rd
@@ -149,12 +149,14 @@ lmer(formula, data = NULL, REML = TRUE, control = lmerControl(),
 \seealso{
   \code{\link[stats]{lm}} for linear models;
   \code{\link{glmer}} for generalized linear; and
-  \code{\link{nlmer}} for nonlinear mixed models.
+  \code{\link{nlmer}} for nonlinear mixed models. \cr
+  \code{\link{plot.merMod}} for plot diagnostics.
 }
 \examples{
 ## linear mixed models - reference values from older code
 (fm1 <- lmer(Reaction ~ Days + (Days | Subject), sleepstudy))
-summary(fm1)# (with its own print method; see class?merMod % ./merMod-class.Rd
+summary(fm1) # (with its own print method; see class?merMod % ./merMod-class.Rd
+plot(fm1) # plotting the model diagnostics; see ?plot.merMod
 
 str(terms(fm1))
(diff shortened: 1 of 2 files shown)


bbolker:
This looks good overall but, unsurprisingly, I have a lot of comments.  General comments here, possibly more specific ones in the commit itself.

* I don't like the `\cr` tags. This is non-semantic markup; I realize that it may not be possible to do proper semantic markup with `Rd` syntax (the natural, semantic thing would be to denote something as a new paragraph rather than a line break ..., or possibly an itemized list if you have a bunch of entries)
* adding a few aliases to improve visibility is OK, but the additions to `plot.merMod.Rd` are overkill. (I would probably stick to just `diagnostics`, `diagnostic_plots` ... seems unlikely that users are going to search in camel-case? I also add `qqplot`, which is probably how people would be searching for a Q-Q plot ...
* "Some of these plots are basic and may be supplemented or superseded by functions from other": delete "are basic and" ? I don't know whether it's possible to add an explicit cross-link to the "See Also" section, instead of explicitly telling users to look there (maybe not; the [Cross-reference section in WRE](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Cross_002dreferences) is about cross-references *between* man pages ...)
* FWIW, I hadn't noticed these [style guidelines](https://developer.r-project.org/Rds.html) referenced in WRE before ...
* I think package names can be marked up as `\CRANpkg{}`

annahuynhly:
1. Got rid of \cr tags and just used itemize.
2. Got rid of the aliases that didn’t start with “diagnostics” and added a qqplot alias.
3. Got rid of “basic”. I spent some time trying to see if it was possible to add an explicit cross-link (I tried earlier), but I couldn’t find any. (I’ll ask on stackoverflow on the meantime.)
4. I see the style guidelines; the only thing I thought was applicable was that “please put two spaces after the end of a sentence in sections with running text”. I looked around the existing documentation and it seems like it’s often used, so I made a slight edit too.
5. Used it; thanks!

Issue #774, opened 2024-03-24
clean up and change default naming args for confint/profile


bbolker:
**tl;dr** propose to (1) make "new"/interpretable names of random-effect parameters the default for `profile`/`confint`, (2) harmonize the argument names across these functions. The main question is: *what is the best adjustment/deprecation path for these backward-incompatible changes*? (jump to *** for discussion)

At present the default behaviour of `confint.merMod` and `profile.merMod` is to use names for the random-effects components `.sig01`, `.sig02`, ... and `.sigma` for the residual variance. This works well with the plotting methods for profile objects &mdash; names of this form get [automatically converted to appropriate `plotmath` expressions] &mdash; but they are inscrutable/non-transparent for users.

About [11 years ago](https://github.com/lme4/lme4/issues/64) we wrote code to produce more informative names (of the form, e.g., `sd_(Intercept)|Subject`). These are requested by calling `confint` with `oldNames = FALSE` *or* `profile` with `signames = FALSE`. There are two problems with this:

* these options are not particularly visible; I suspect most users don't know about them
* it's a little weird to have two different argument names doing essentially the same thing in two functions (*and* using different naming conventions; the horse is way out of the barn, but it would be good to try to define what we *thought* the naming conventions were for `lme4`, and gradually adjust things internally and [slower] externally).

We could start by issuing a warning if the `signames`/`oldNames` arguments weren't specified explicitly (i.e. by testing `missing()`).  This would be annoying if repeated frequently. Depending on how fancy we want to be, we could warn once per session or every xx hours [see machinery in `lifecycle::deprecate_warn` &mdash; I'm nervous about depending on `lifecycle` since it's tidyverse-ish, but it only has three non-base reverse dependencies (`cli`, `glue`, `rlang` ...].

We could simultaneously add new arguments (or change one function to be the same as the other) while keeping the old argument in place to allow the warning, e.g. if both functions used `oldnames` or `old.names` (either of which would be more consistent with the dominant argument naming convention), we could warn on use of `signames`/`oldNames` ...

---

PS This also exposes a bug/infelicity, namely that `confint()` inherits the naming setting from `profile`, even if it's explicitly overridden ...

```r
library(lme4)
fm1 <- lmer(Reaction ~ Days + (Days|Subject), sleepstudy)
pp <- profile(fm1, signames = FALSE)
ci <- confint(pp, oldNames = TRUE)
rownames(ci)
[...]

PR #837, opened 2025-07-03
Changing the oldNames argument in confint to be consistent with profile

annahuynhly:
This pertains to this issue: https://github.com/lme4/lme4/issues/774 

In summary, what I did was use signames for both, added a warning that oldNames will be depreciated, changed the documentation to account for the new argument. My rationale:

1. Personally I think "signames" is better than "oldNames" because it provides a better description of what form the names look like.
2.  Using periods such as “sig.names” or “old.names” might change too much. This doesn’t really seem that standardized across the board either; some people prefer underscores.
3. Should probably provide a warning to give people time to adjust, and also, for their code to run. Probably should remove the warning message in a couple years.

### DIFF
--- a/R/profile.R
+++ b/R/profile.R
@@ -796,23 +796,27 @@ confint.thpr <- function(object, parm, level = 0.95, zeta,
 ##' @param nsim number of simulations for parametric bootstrap intervals
 ##' @param boot.type bootstrap confidence interval type
 ##' @param quiet (logical) suppress messages about computationally intensive profiling?
-##' @param oldNames (logical) use old-style names for \code{method="profile"}? (See \code{signames} argument to \code{\link{profile}}
+##' @param signames (logical) use old-style names for \code{method="profile"}? (See \code{signames} argument to \code{\link{profile}}
 ##' @param \dots additional parameters to be passed to  \code{\link{profile.merMod}} or \code{\link{bootMer}}
 ##' @return a numeric table of confidence intervals
 confint.merMod <- function(object, parm, level = 0.95,
                            method = c("profile","Wald","boot"),
                            zeta, nsim=500, boot.type = c("perc","basic","norm"),
-                           FUN = NULL, quiet=FALSE, oldNames=TRUE, ...)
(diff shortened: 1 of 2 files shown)


bbolker (commented):
Could you also please add an item to `inst/NEWS.Rd` describing this change?

bbolker:
I'm not too worried about the date on NEWS (almost tempted to replace it by "current", but the format is fussy about date specifications)

jaganmn:
@annahuynhly Are you able to fix this?

```
* checking for code/documentation mismatches ... WARNING
Codoc mismatches from Rd file 'confint.merMod.Rd':
confint.merMod
  Code: function(object, parm, level = 0.95, method = c("profile",
                 "Wald", "boot"), zeta, nsim = 500, boot.type =
                 c("perc", "basic", "norm"), FUN = NULL, quiet = FALSE,
                 oldNames = NULL, signames = TRUE, ...)
  Docs: function(object, parm, level = 0.95, method = c("profile",
                 "Wald", "boot"), zeta, nsim = 500, boot.type =
                 c("perc", "basic", "norm"), FUN = NULL, quiet = FALSE,
                 oldNames = TRUE, ...)
  Argument names in code not in docs:
    signames
  Mismatches in argument names:
    Position: 11 Code: signames Docs: ...
  Mismatches in argument default values:
    Name: 'oldNames' Code: NULL Docs: TRUE

* checking Rd \usage sections ... WARNING
Undocumented arguments in Rd file 'confint.merMod.Rd'
  ‘oldNames’
Documented arguments not in \usage in Rd file 'confint.merMod.Rd':
  ‘signames’

Functions with \usage entries need to have the appropriate \alias
entries, and all their arguments documented.
The \usage entries must correspond to syntactically valid R code.
See chapter 'Writing R documentation files' in the 'Writing R
Extensions' manual.
```

The two issues are:
* You added `signames` to the code but that is not reflected in the `\usage{}` of `confint.merMod.Rd`.
* You removed `oldNames` from the `\arguments{}` of `confint.merMod.Rd`, but it has to remain there until `oldNames` is actually removed from the code.

You'll know that they are fixed if `R CMD check` no longer produces the above output.

Issue #782, opened 2024-04-25
check consistency of docs, code etc for Gamma dispersion parameter


lweasel:
_(Please bear in mind that I very likely might have got the wrong end of the stick here!)_

I'm trying to simulate data from a model fitted with ```glmer()``` using the Gamma family and ```simulate.merMod()```; however the simulated values I get are noticeably different in distribution to the original data with which the model was fitted. On the other hand, if I ignore random effects, fit the original data using ```stats::glm()```, then simulate data from that fitted model, then the simulated and original values look far more similar.

I noticed that in commit a46fcef the change was made when calculating the Gamma shape parameter for simulation from using ```MASS::gamma.shape(model)$alpha``` to ```sigma(model)``` (I think that the former is used when simulating data from a Gamma model fitted with ```stats::glm()``` - but wouldn't work for a merMod object?).

I also noticed that issue #291 says that the Gamma shape parameter for a model is accessible as ```1/sigma(model)^2```. So I wondered if the use of ```sigma(model)``` in ```Gamma_simfun()``` in predict.R is right?

(p.s. in this test case, my original data was actually hand-simulated from a Gamma distribution, and ```1/sigma(model)^2``` gives a fitted shape parameter very close to the value I originally used for hand-simulation).

bbolker:
You're exactly right.

```r
dd <- expand.grid(group = 1:50, rep = 1:50)
set.seed(101)
b <- rnorm(50)
eta <- 1 + b
mu <- exp(eta)
y <- rgamma(nrow(dd), shape = 2, scale = mu/2)
dd2 <- data.frame(dd, y)
library(lme4)
m1 <- glmer(y ~ 1 + (1|group), family = Gamma(link = "log"), data = dd2)
ss <- simulate(m1, nsim = 1000)
res <- sapply(ss, function(x) c(mean(x), sd(x)))
par(mfrow = c(1,2))
hist(res[1,], main = "mean", breaks = 50)
abline(v=mean(dd2$y), col = 2, lwd = 2)
hist(res[2,], main = "sd", breaks = 50)
abline(v=sd(dd2$y), col = 4, lwd = 2)

hist(log(dd2$y), freq = FALSE, breaks = 50)
cc <- adjustcolor("black", alpha.f = 0.1)
invisible(lapply(ss, \(x) lines(density(log(x)), col = cc)))
```

Issue #291, opened 2015-03-30
Gamma shape parameter information not printed/documented


bbolker:
The shape parameter for Gamma models is accessible as `1/sigma(fitted_model)^2`, but this is not printed in the summary nor documented anywhere obvious.

PR #841, opened 2025-07-15
Adding a Test and Fixing Documentation for Gamma GLMMs

annahuynhly:
This issue (https://github.com/lme4/lme4/issues/782) details a bug regarding the Gamma dispersion parameter, which has been fixed. Another issue (https://github.com/lme4/lme4/issues/291) discusses how the documentation for simulate.merMod may be incorrect (which it was); and I added that the shape parameter can be found via `1/sigma(fitted_model)^2` instead.

To check for future bugs, I also added a test to ensure we'd get a shape parameter from a particular simulation.

I used `tolerance = 0.05` since after simulating for `100` times, the average distance was around `0.05`. Simulation of the decision is shown below:

```
library(lme4)

n_runs <- 100
shape_vals <- dist_vals <- rep(0, length.out = n_runs)
dd <- expand.grid(group = 1:50, rep = 1:50)
shape_gam <- 2

set.seed(101)
for(i in 1:n_runs){
  b <- rnorm(50)
  eta <- 1 + b
  mu <- exp(eta)
  y <- rgamma(nrow(dd), shape = shape_gam, scale = mu/2)
  dd2 <- data.frame(dd, y)
  
  m1 <- glmer(y ~ 1 + (1|group), family = Gamma(link = "log"), data = dd2)

  shape_vals[i] <- 1/sigma(m1)^2
  dist_vals[i] <- abs(shape_gam - 1/sigma(m1)^2)
}
```

Then `mean(dist_vals)` gave `0.05410642`.


bbolker:
I would suggest organizing the test a little bit more like this:

* test that the `simulate.formula()` method is behaving the same as a more transparent but less general simulation function
* test a *single* run (with specified seed) at tolerance = 0.05
* test against that specific value (=1.945 ...) with a much tighter tolerance (allowing for among-platform variation), to flag future changes/breakage

(For inclusion in tests, replace `all.equal()` with `testthat::all_equal()` ...)

```r
library(lme4)
simfun_gam <- function(ngrp = 50, nrep = 50, shape_gam = 2, intercept = 1, use_simulate = FALSE, seed = NULL) {
   if (!is.null(seed)) set.seed(seed)
   dd <- expand.grid(group = 1:ngrp, rep = 1:nrep)
   if (use_simulate) {
       dd$y <- simulate(~ 1 + (1 | group), newdata = dd, family = Gamma(link = "log"), 
                 newparams = list(theta = 1, beta = 1, sigma = 1/sqrt(shape_gam)))[[1]]
       return(dd) 
   }
   b <- rnorm(ngrp)
   eta <- intercept + b  ## 
   mu <- exp(eta)
   y <- rgamma(nrow(dd), shape = shape_gam, scale = mu/shape_gam)
   data.frame(dd, y)
}
dd1 <- simfun_gam(seed = 101)
dd2 <- simfun_gam(seed = 101, use_simulate = TRUE)
all.equal(dd1$y, dd2$y)
m1 <- glmer(y ~ 1 + (1|group), family = Gamma(link = "log"), data = dd1)
all.equal(1/sigma(m1)^2, 2.0, tolerance = 0.05)
all.equal(1/sigma(m1)^2, 1.94511502080571, tolerance = 1e-6)
```

Issue #724, opened 2023-04-22
allFit improvements


bbolker:
It should instead use the subset of optimizers that are available.

bbolker:
The original problem is fixed.

- plot method?
- better behaviour/documentation around what "OK"/"failure" means (complete failure vs warning vs ...)

bbolker:
It should allow an option of starting from the values estimated by the initial fit (this will presumably improve performance for large/slow models ...)

PR #845, opened 2025-07-23
allFit Update to Allow Initialization from Model's MLE

annahuynhly:
allFit Update to Allow Initialization from Model's MLE

This pull request is related to issue https://github.com/lme4/lme4/issues/724.

In summary: added a new argument `start_from_mle` to `allFit` so the user may initialize 
the refitting process using starting values from the model’s MLE. 

Example of use:
```
gm1 <- glmer(cbind(incidence, size - incidence) ~ period + (1 | herd),
             data = cbpp, family = binomial)
## show available methods
allFit(show.meth.tab=TRUE) 
gm_all1 <- allFit(gm1)
ss1 <- summary(gm_all1)

gm_all2 <- allFit(gm1, start_from_mle = TRUE)
ss2 <- summary(gm_all2)

ss1
ss2
```

Extra:

1. I made minor changes in the documentation in `allFit()` to be a bit more consistent. I.e., I noticed `logical: ...` and `(logical) ...` are used. Switched to ensure all followed the former.
2.  I checked to see if `allFit()` works fine on a Window machine (see issue https://github.com/lme4/lme4/issues/554) and things are working fine, including the version in this pull request!

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -15,7 +15,9 @@ name: R-CMD-check
 
 jobs:
   R-CMD-check:
-    if: "contains(github.event.head_commit.message, '[run ci]')"
+    if: |
+      github.event.name == "pull_request" ||
+       contains(github.event.head_commit.message, '[run ci]')
 
     runs-on: ${{ matrix.config.os }}
 
@@ -35,6 +37,7 @@ jobs:
       R_REMOTES_NO_ERRORS_FROM_WARNINGS: true
       RSPM: ${{ matrix.config.rspm }}
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+      LME4_TEST_LEVEL: 2
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -15,8 +15,7 @@ name: R-CMD-check
 
 jobs:
   R-CMD-check:
-    if: |
-      github.event.name == "pull_request" ||
+    if: github.event.name == 'pull_request' ||
        contains(github.event.head_commit.message, '[run ci]')
 
     runs-on: ${{ matrix.config.os }}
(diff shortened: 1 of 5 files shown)


bbolker:
as usual, looking at this carefully reminds me of a few other things, which may or may not need to be done now.

* I tried a quick benchmark (see commented code near the top of `tests/testthat/test-allFit.R`), which didn't find any difference. I can think of a few reasons:
     * the machinery is actually not working as we want it to (i.e., for some reason the new 'start' specification isn't kicking in/being copied where it needs to)
     * the fitting procedure is too easy/short to see a difference
     * for GLMMs there is a complication/additional step: by default, GLMMs run a two-stage optimization procedure, the first time with `nAGQ=0` [a fast approximation], and then with `nAGQ>1`, *unless* (1) the user opts for using only the approximation (setting `nAGQ=0` in the `glmer` call) or (2) opts to skip the initial step (`control = glmerControl(nAGQ0initStep = FALSE)`. There should probably be a way to control this, and to turn off the initial step, when running a GLMM with specified starting values (?? unless this is already there in the code ?? it might be ... it could be a sensible general default, although now that I'm thinking about it I can of course think of some more complications ...). 
     
Probably the next thing to try in testing is to try `allFit` for a large-ish LMM, e.g.

```r
ifit <- lmer(y ~ studage + lectage + (1|s) + (1|d), data = InstEval)
```

(this is 73K rows and takes 3 seconds to fit, but only 34 function evaluations ...) and see what kind of difference `start_from_mle` makes ...


More generally it would be nice to have a way to insert/override control statements, e.g. set `calc.derivs = FALSE` if we're not interested in seeing which fits have convergence warnings but only what the estimates are ...

It occurs to me that checking the number of function evaluations required is slightly more reliable than time, although that too seems to be mixed (at least the *average* number of function evaluations is higher for 'nostart')

```r
 get_feval <- function(x) x@optinfo$feval
cbind(start = sapply(gm_all, get_feval), nostart = sapply(gm_all_nostart, get_feval))
bobyqa                           59      73
Nelder_Mead                     176     323
nlminbwrap                       NA      NA
nmkbw                           187     180
optimx.L-BFGS-B                   8      18
nloptwrap.NLOPT_LN_NELDERMEAD   184     178
nloptwrap.NLOPT_LN_BOBYQA        39      53
```

bbolker:
@jaganm -- probably ...

PR #848, opened 2025-07-23
code/documentation mismatch from changing oldNames argument (R CMD check issue)

annahuynhly:
@jaganmn pointed out an issue here: https://github.com/lme4/lme4/pull/837#issuecomment-3104731223 

There were a few things I forgot to modify when making changes... In this newest version there should no longer be no further issues pertaining to the oldNames/signames problem. :)))


annahuynhly:
Thanks for fixing remaining issues and letting me know a command to make it easier for me in the future; pretty sure you could tell I'm new to this haha!

Issue #238, opened 2014-08-27
add GLMM data sets


bbolker:
There are several more data sets that we might dig out/add to the package for general use/testing:
- _Culcita_/coral data set, gopher tortoise data set (from http://rpubs.com/bbolker/glmmchapter )
- Banta 2010 _Arabidopsis_ fertilization/clipping responses (used as an example by Bolker et al. GLMM review 2009)
- the classic 'toenail' data, available from http://www.stata-press.com/data/r12/xt.html
- salamander cross data, available e.g. from `HGLMM` or `spaMM` packages
- schizophrenia data from Hedeker and Gibbons referred to by http://comments.gmane.org/gmane.comp.lang.r.lme4.devel/12135
- perhaps an extended version of the CBPP data set, including more covariate information (from Bitbucket repos for glmer JSS paper)

PR #855, opened 2025-07-30
Adding extra datasets

annahuynhly:
This pull request pertains to this issue: https://github.com/lme4/lme4/issues/238

I also wrote documentation for the added datasets.

Main issue is that I did not explicitly ask for permission; but all of these were publicly available; some of them were from previous packages, and the rest were from Ben. :)


bbolker:
Thanks, I'll take a look.

A few more things:

* it's best to use `rds` rather than `rda` if you're saving a single R object -- `rda` stores (1 or more) *named* objects, which is often less convenient (you have to save the return value to figure out what named objects were actually in the file ...)
* you should make sure to use `version=2` when saving, otherwise we will have to bump the required R version to something higher (4.0?). (To be honest, this level of backward compatibility is probably nearly a lost cause at this point, but I like to maintain back-compatibility whenever it's reasonably easy to do so ...) [did you run `R CMD check`, or `devtools::check()`, on this version? I tried to get GitHub actions to check pull requests automatically, but don't seem to have managed it ...]
* it might be for a separate issue, but I'd like to consider starting to use the `Rdpack` package so that we could maintain references in BibTeX format - see [here](https://cran.r-project.org/web/packages/Rdpack/vignettes/Inserting_bibtex_references.pdf). It does add one more dependency, but `Rdpack` has only one non-default dependency itself, so it's pretty mild ...

annahuynhly:
Good to know; most of the current datasets are in the `.rds` format instead of the `.rda` format, so I wanted to be consistent. I will look into using `.rds` instead (Is this something we’d also want to modify)?

I believe I didn’t use `version=2`; I’ll modify this. I did not run `R CMD check` or `devtools::check()` so I’ll also try that and update when you are free to check again.

Agree with considering to use the `Rdpack` package; I just don’t know if you want me to do it right now due to the dependencies, but let me know if that's another potential project in the works...

bbolker:
Apparently we already depend indirectly on `Rdpack`, at least in the development version, via `reformulas`.  It's kind of overkill for `reformulas` (which only uses one citation!), but ...

In any case: go ahead and start implementing this (citations via `Rdpack`) if you like.

```
pd <- tools::package_dependencies
"Rdpack" %in% pd("lme4", recursive=TRUE)$lme4  ## TRUE
sapply(pd("lme4")$lme4, function(x) "Rdpack" %in% pd(x)[[1]])  ## reformulas is the culprit
```

annahuynhly:
I was going crazy wondering why I couldn't commit the `.rda` files. I checked .gitignore and behold:

```
(base) Annas-Air:lme4 annaly$ cat .gitignore
TAGS
*~
\#*
.#*
*.gz
src/*.o
src/*.so
src/*.dll
R/*.R+*
vignettes/*.R
vignettes/*.tex
vignettes/cache/*
inst/doc/cache/*
inst/testdata/lme-tst-fits.Rout*
inst/testdata/*testdata*
tests/*.Rout*
tests/*.R+*
tests/testthat/*.R+*
.Rproj.user
lme4.Rproj
.Rhistory
misc/
revdep/
.gitsum/
Rplots.pdf
*.rds
## docs/
## built on GH from inst/NEWS.Rd
## NEWS.md
```

So... Was there a good reason in the past? Can I just get rid of this?

bbolker:
Use git add -f

annahuynhly:
Just double checking that you explicitly wanted this as a feature.

So for most of the other files (before the ones I added) which were explicitly `.rda` files, we can easily load them in instantly. For instance,
```
> devtools::load_all()
 Loading lme4
> head(sleepstudy)
  Reaction Days Subject
1 249.5600    0     308
2 258.7047    1     308
3 250.8006    2     308
4 321.4398    3     308
5 356.8519    4     308
6 414.6901    5     308
> head(cbpp)
  herd incidence size period
1    1         2   14      1
2    1         3   12      2
3    1         4    9      3
4    1         0    5      4
5    2         3   22      1
6    2         1   18      2
```
However, if I do this with the `.rds` files then,
```
> head(cbpp2)
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'x' in selecting a method for function 'head': object 'cbpp2' not found
```
If I load the file like so, all is okay:
```
> cbpp2 <- readRDS("data/cbpp2.rds")
> head(cbpp2)
  herd treatment avg_size period size incidence
1    1  Complete     23.7      1   22         8
2    1  Complete     23.7      2   16         2
3    1  Complete     23.7      3   16         0
4    1  Complete     23.7      4   20         2
5    2  Complete     19.0      1   17         3
6    2  Complete     19.0      2   17         0
```
Although, I imagine using the directory `"data/cbpp2.rds"` would be quite inconvenient... Before I was just following the [R Packages](https://r-pkgs.org/data.html#sec-data-data) textbook, where they generally recommended `.rda` files for packages. I tried to find other resources for using `.rds` in packages and I didn't find much.

bbolker:
Hi @annahuynhly : I'm very sorry, but you were right in the first place and I was wrong about `.rda` vs. `.rds` files. For most purposes I stand by my advice about preferring `.rds` to `.rda`, but for the specific purpose of including data sets in R packages `.rda` is actually better, for the reasons you discovered ... I hope it won't be too much trouble to switch them back (still using `version=2` though) ...

PR #856, opened 2025-07-31
Man page for summary.merMod

annahuynhly:
This pull request is associated with this comment: https://github.com/lme4/lme4/issues/291#issuecomment-3120182144  

Essentially,
- Made the summary.merMod man page, referenced sigma.merMod,
- Made VarCorr.merMod reference sigma.merMod,
- Updated sigma.merMod to also reference what it means for poisson and binomial fitted LMMs. Couldn’t really find out what it meant for inverse gaussian, and quasi families are not supported so did not describe those families.


bbolker:
in general the dispersion parameter `sigma` is the *square root* of the constant multiplier in the 'variance function' provided by the `family()` functions, however that translates to the usual parameterization of the distribution. e.g. for Gaussian: `gaussian()$variance` is 1 (more precisely `rep.int(1, length(mu))`), so `sigma` is $\sqrt{\sigma^2}} = \sigma$.

For Gamma,

- Variance = `shape*scale^2`
- `Gamma()$variance`: $\mu^2$. 

So `sigma` is the square root of the **shape** parameter (I think the current version might say "square root of the **dispersion** parameter"; if so, please edit ...)

For inverse Gaussian, Variance = $\mu^3/\lambda$ (from [Wikipedia definition](https://en.wikipedia.org/wiki/Inverse_Gaussian_distribution), and `inverse.gaussian()$variance` is `mu^3`, so `sigma()` would be the square root of the *reciprocal* of the shape parameter (might be clearest to say "$\sqrt{\lambda}$, where $\lambda$ is the shape parameter).  $\lambda$ is the scaled drift parameter for the first-passage time of a Brownian motion, but that's probably going too far down the rabbit hole for this man page ...

annahuynhly:
For the inverse Gaussian case, I was thinking of `sqrt(lambda)` at some point, but I generated some data and noticed the value was (in my opinion) very different, so I was skeptical. Maybe this could be the fault of how I tried to generate the data:

```
library(lme4)

set.seed(123)

n <- 5000
n_groups <- 50

x_vec <- rnorm(n, mean = 0, sd = 1)  
group <- factor(rep(1:n_groups, each = (n/n_groups))) 

mu <- rnorm(n, mean = 2, sd = 0.5)
lambda <- 20
y <- statmod::rinvgauss(n, mean = mu, dispersion = lambda)

df <- data.frame(y, x_vec, group)

gm1 <- glmer(y ~ x_vec + (1 | group), family = inverse.gaussian(link = "log"), 
             data = df)
```
Then we get:
```
> sigma(gm1)
[1] 3.237415
> sqrt(lambda)
[1] 4.472136
```


Anyhow, I'll definitely edit the rest of the explanations.

annahuynhly:
> @annahuynhly Did you want to make any further changes or should I merge this?

I'm going to make some more changes tonight... I am working on a bunch of meal prep after our meeting lol.

annahuynhly:
Oops... My bad; I wasn't done. There are some problems with `R CMD check` (I was just working on them...)

I'll just do a new pull request once I fix all of the issues...

Issue #853, opened 2025-07-29
make sure to restore formula environment contents after calling allFit


bbolker:
we assign `pars` and `ctrl` (maybe?) to the formula environment inside `allFit` so that we can `update`/`eval` things safely.

For maximum safety, we should record the existing values in the environment (probably `NULL`) before assigning, and use `on.exit()` to restore when leaving the function ...

e.g. see [here](https://github.com/glmmTMB/glmmTMB/blob/master/glmmTMB/R/predict.R#L237-L258) in `glmmTMB` for an example ...

HenrikBengtsson:
Yes, it's certainly better than currently. Though, one can come up with corner cases where this does not work. For example, with:

```r
delayedAssign("ctrl", stop("boom"))
```

you'd trigger an error if you try to temporarily record the original `ctrl`, e.g. `ctrl_org <- ctrl`. Of course, that's rare.


If we go back to what @bbolker said:

> we assign pars and ctrl (maybe?) to the formula environment inside allFit so that we can update/eval things safely.

So, if the above set-and-undo strategy will only apply inside of `allFit()`, then you are in full control, have you checked that you really have to attach those objects in the formula?  For instance, if I look at `update()` in Ben's comment, you already pass `ctrl` as an explicit argument;

https://github.com/lme4/lme4/blob/65cc514c1f4a946ae3fe8db75912198df3534b6c/R/allFit.R#L183-L187

My point is, could the assignment of `pars` and `ctrl` be a left over from the past that you do not really need here?

Disclaimer: I don't have a deep understanding of **lme4**.

HenrikBengtsson:
Also, you might be able to `eval()` via a temporary environment holding those objects. Gist:

```
form <- formula(object)
env <- environment(form)
tmp_env <- new.env(parent = env)
assign("pars", pars, envir = tmp_env, inherits = FALSE)
assign("ctrl", ctrl, envir = tmp_env, inherits = FALSE)
environment(form) <- tmp_env
on.exit({ environment(form) <- env }, add = TRUE)
```

PR #870, opened 2025-10-01
Preventing environment leaks for allFit

annahuynhly:
This pull request pertains to the following issue: https://github.com/lme4/lme4/issues/853

I used Henrik Bengtsson's suggestion for evaluating and assigning `pars` and `ctrl` within a temporary environment to prevent potential issues.

Also, modified some of the tests to ensure there's no future leakage, as well as added a line to one of the existing tests (which caused some errors earlier).

### DIFF
--- a/tests/testthat/test-allFit.R
+++ b/tests/testthat/test-allFit.R
@@ -4,12 +4,17 @@ if (testLevel>1) {
   L <- load(system.file("testdata", "lme-tst-fits.rda",
                         package="lme4", mustWork=TRUE))
   
+  had_pars <- exists("pars", envir = globalenv(), inherits = FALSE)
+  had_ctrl <- exists("ctrl", envir = globalenv(), inherits = FALSE)
+  
   gm_all <- allFit(fit_cbpp_1, verbose=TRUE)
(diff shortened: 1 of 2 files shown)


tests/testthat/test-allFit.R
```diff
@@ -7,6 +7,11 @@ if (testLevel>1) {
   gm_all <- allFit(fit_cbpp_1, verbose=TRUE)
   gm_all_nostart <- allFit(fit_cbpp_1, verbose=FALSE, start_from_mle = FALSE)
 
+  test_that("ensuring global environment is not contaminated", {
+    expect_equal(!("pars" %in% ls()), TRUE)
```
bbolker: I think these could be `expect_false("pars" %in% ls())` or `expect_false(exists("pars"))`

annahuynhly: I just realized that you looked at outdated code... I changed the test quite a bit in the latest committed version, i.e., I did something like:

`expect_equal(exists("pars", envir = globalenv(), inherits = FALSE), had_pars)`
                 
 Instead, since we went for the temporary environment solution...

Issue #783, opened 2024-04-26
lme4 and optimisers disagree on convergence


HeikoSchuett:
Hi! 
I am quite often running into warnings for lme4 models where it claims something like the following:
```
optimizer (nloptwrap) convergence code: 0 (OK)
Model failed to converge with max|grad| = 0.00229598 (tol = 0.002, component 1)

```
, i.e. the optimiser believes the optimisation converged fine, but lme4 thinks that the gradient is (slightly) too large and raises a warning. 

After looking at the optimisation reports in a bit more detail the usual problem is that the optimiser terminates because the steps in  parameter values are small, while the gradient is not yet close to 0, i.e. for large Hessians. Telling the optimiser to ignore this (`control = lmerControl(optCtrl = list(xtol_abs=0, xtol_rel=0)))`) does run the optimiser for a few steps more and yields virtually identical results. Thus, I am fairly certain that nothing here is actually problematic, but apparently the warning criteria of lme4 are substantially different from the ones used by the optimiser(s) such that the optimiser does not do the last few steps to converge to lme4s satisfaction.

However, I am teaching a course on mixed models using R and error messages like this are incomprehensible to my psychology students and very unsettling for them,  because they have no idea how to fix this and are uncertain about any of the results they get.

The only alternative solution I found is to normalise the output and/or the predictors. This complicates the interpretation of the coefficients, which I would like to avoid unless it is really necessary. For the examples I am using, the means and standard-deviations tend to be single or at most 2 digit numbers, such that I don't think the scale causes any actual numerics problems.

Is there a chance to homogenise the criteria of the checks done by the optimiser and the ones done by lme4? This would certainly avoid a lot of confusion in my class and most probably for other researchers, too. 

best
Heiko

P.S: a simple script that creates the problem for me (R= 4.3.3, lme4=1.1-35.2) is this generated data I intend to use for an exercise:
```
set.seed(4)

data <- data.frame(row.names = 1:150)
data$class <- sample(1:25, 150, replace = TRUE)
data$grade <- pmin(7 + floor(runif(150,2,14) + rnorm(150)), 20)
data$emint  <- rnorm(150, 100, 15)
data$group <- rep(c("training", "control", "school"), each=50)

data$rater <- sample(1:10, 150, replace = TRUE)
rater_bias <- rnorm(10, 0, 2)

training_effect <- list(training=2.5,school=2,control=0)

class_bias <- rnorm(25, 0, 1)

emint_effect <- 0.001
[...]

PR #871, opened 2025-10-02
Adding conv.nparmax to merControl()

annahuynhly:
Refer to the following comment: https://github.com/lme4/lme4/issues/783#issuecomment-3156476503 

Also added:
- Test to ensure that `conv.nparmax` works
- Updated news and relevant documentation (did run `R CMD check`)

Small concerns:
- Set the default to be `=50`; unsure if this is too large (or too small!)
- The example for the test isn't the most elegant; had a hard time reproducing an issue in a more efficient manner


bbolker:
`theta` > 10 seems reasonable (since most of the uses of the Hessian are bogus when the fit is singular, which is likely in this case ...). Not entirely clear what the criterion on `length(theta) + length(beta)`, which is relevant in the GLMM case, should be ... maybe `length(theta)>10 || length(theta) + length(beta) > 20` ?

annahuynhly:
@bbolker probably all good now...

Issue #866, opened 2025-08-27
`predict()` fails for `glmer()` when `se.fit = TRUE`


strengejacke:
``` r
set.seed(123)
dat <- data.frame(
  outcome = rbinom(n = 100, size = 1, prob = 0.35),
  var_binom = as.factor(rbinom(n = 100, size = 1, prob = 0.7)),
  var_cont = rnorm(n = 100, mean = 10, sd = 7),
  grp = as.factor(sample(letters[1:4], size = 100, replace = TRUE))
)
m1 <- lme4::glmer(
  outcome ~ var_binom + var_cont + (1 | grp),
  data = dat,
  family = binomial(link = "logit")
)
d <- insight::get_datagrid(m1, "var_binom", include_random = TRUE)
d
#>   var_binom var_cont grp
#> 1         0 9.247172   a
#> 2         1 9.247172   a

predict(m1, newdata = d, se.fit = TRUE)
#> Warning in predict.merMod(m1, newdata = d, se.fit = TRUE): se.fit computation
#> uses an approximation to estimate the sampling distribution of the parameters
#> Error in h(simpleError(msg, call)): error in evaluating the argument 'x' in selecting a method for function 'rowSums': non-conformable arguments
```

<sup>Created on 2025-08-27 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

bbolker:
I don't think this is actually `glmer`-specific?

```r
m2 <- lme4::lmer(outcome ~ var_binom + var_cont + (1 | grp), data = dat)
predict(m2, newdata=d, se.fit= TRUE)
```

fails in the same way.

Fails in `quad.tdiag(Cmat, ZX)` with these objects:

```
str(Cmat)
Formal class 'dpoMatrix' [package "Matrix"] with 5 slots
  ..@ Dim     : int [1:2] 7 7
  ..@ Dimnames:List of 2
  .. ..$ : chr [1:7] "grp.a.(Intercept)" "grp.b.(Intercept)" "grp.c.(Intercept)" "grp.d.(Intercept)" ...
  .. ..$ : chr [1:7] "grp.a.(Intercept)" "grp.b.(Intercept)" "grp.c.(Intercept)" "grp.d.(Intercept)" ...
  ..@ x       : num [1:49] 0.0819 0 0 0 0 ...
  ..@ uplo    : chr "U"
  ..@ factors : list()


Browse[1]> str(ZX)
[...]

PR #874, opened 2025-10-13
Subsetting cmat

annahuynhly:
This pull request is to fix the following issue outlined here: https://github.com/lme4/lme4/issues/866

Some details of the problem is outlined in the issue; to resolve these issues, `Cmat` from `predict.merMod()` has been subsetted to only include the groups that's present in the `newdata` argument.

A test has also been added to check if there are any errors running codes with `newdata` that do not incorporate all levels of a group.

### DIFF
--- a/R/checkConv.R
+++ b/R/checkConv.R
@@ -16,7 +16,10 @@ help_str <- "\n  See ?lme4::convergence and ?lme4::troubleshooting."
 ##'   (length is taken to determine number of RE parameters)
 ##' @param debug useful if some checks are on "ignore", but would "trigger"
 ##' @param nobs the number of observations from the dataset
-checkConv <- function(derivs, coefs, ctrl, lbound, debug = FALSE, nobs = NULL)
+##' @param ndim the number of dimensions for the variance-covariance matrix 
+##' of random effects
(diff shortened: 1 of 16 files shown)


bbolker:
I think my new code is an improvement, but I can think of ways to break it ... what if we have two different groups with the same sets of labels but different subsets of labels included ... ????

annahuynhly:
After I've pulled your changes, it appears that there are some test cases that fail (and of course my new modified version fails too as a consequence)... I'm going to commit my changes anyways and see if the test cases were even reasonable to begin with...

annahuynhly:
@bbolker  also should be good for merging now.

annahuynhly:
Yeeep sadly I noticed that once we have a non-intercept model, the extra code I have does not work for subsetting `Cmat` properly. However, neither does the original code either. I may ask you more questions next week on what Cmat is supposed to actually look like; I'mma take a big gander at the documentation though...

Another example where `Cmat` doesn't match the dimensions of `ZX`:
```
library(lme4)
m1 <- lmer(Reaction ~ Days + (1 + Days | Subject), data = sleepstudy)

set.seed(1)
d <- sleepstudy[sample(1:nrow(sleepstudy), size = 30), ]

predict(m1, newdata = d, se.fit = TRUE)
```

PR #876, opened 2025-10-20
Improved Cmat

annahuynhly:
A follow-up from a previous pull request: [https://github.com/lme4/lme4/pull/874](https://github.com/lme4/lme4/pull/874#event-20361641600)

I had a suspicion that the following issue would take more work: https://github.com/lme4/lme4/issues/753

Turns out the fix was simple. :'^)

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -567,7 +567,8 @@ predict.merMod <- function(object, newdata=NULL, newparams=NULL,
       mask <- unlist(lapply(
         intersect(names(object@flist), names(Z_factors)),
         function(grp) {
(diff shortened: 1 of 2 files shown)


annahuynhly:
The issue was actually **both**; the previous PR only solved the issue where the `newdata` only has a smaller set of levels per group. 

For example, the previous pull request fixed this issue: if the dataset used to fit the model had a group `grp` with 4 levels `a`, `b`, `c`, `d` but the `newdata` had only had levels `a` and `b`, then that causes an error. However, one issue was that it only worked if you had intercept random effects

PR #877, opened 2025-10-24
flexSigma rewrite

jaganmn:
I'd like it if we started to work on this branch, which is a *minimal* version of #849 with somewhat higher code quality.  After some hours of careful code and documentation reading, I determined that the original PR added so much spurious code that debugging and maintaining it would really have become a nightmare.

I'll be happy to meet to review `git diff master` which is here much smaller.  Otherwise, a good place to start is `TODO.md` and working through the `FIXME` comments added under `man/` and `tests/`, which mark places where failing examples and tests have been skipped.  Some tests fail due to small numerical differences; I'm hoping that these can be understood here more easily.

Once the existing examples and tests are sorted, I'll add the code needed to support `cs` and `ar1` and `hom=TRUE`, taking what I can from Greg's work.

I see some big architectural issues:
* `pp` does not store `par` which now can be not identical to `theta`.  Should it?
* We store `reCovs` (a list of `Covariance` objects) as an attribute of the model object.  Methods find it and do stuff with it.  How should methods handle model objects without the attribute, presumably generated by earlier **lme4** which did not set it?
* Should **reformulas** deprecate `mkReTerms(calc.lambdat=TRUE)`, which we no longer use?  I guess not, if new **reformulas** must support old **lme4** ... ?

### DIFF
--- a/R/AllClass.R
+++ b/R/AllClass.R
@@ -1,6 +1,9 @@
 ## pro tem:
-## $ LME4_WORK_AROUND_BLME=true     R CMD INSTALL blme_1.0-6.tar.gz
-## $ LME4_WORK_AROUND_LMERTEST=true R CMD INSTALL lmerTest_3.1-3.tar.gz
+## e.g., for 'blme':
+## $ LME4_WORK_AROUND_BLME=false R CMD INSTALL lme4_2.0.0.tar.gz
+## $ R CMD check blme_1.0-6.tar.gz # broken
+## $ LME4_WORK_AROUND_BLME=true  R CMD INSTALL lme4_2.0.0.tar.gz
+## $ R CMD check blme_1.0-6.tar.gz # ok
 .lme4.work.around.blme <-
     isTRUE(as.logical(Sys.getenv("LME4_WORK_AROUND_BLME"    , "false")))
 .lme4.work.around.lmerTest <-


bbolker:
Thank you, I'll look at this as soon as I can.

None of the architectural issues seems too worrying to me, although maybe that's because I haven't looked carefully yet.

* [`pp` not storing `par`] I'm not sure; not necessarily. (IMO the original design of `lme4` relies too heavily on passing information in environments, which is useful for memory management but also reduces transparency ...)
* [missing attributes] a NULL value for this attribute, or a component of it, is a signal that we should be assuming an unstructured pos-def covariance matrix, no theta transformation, etc. (i.e. old-`lme4`)
* I wouldn't be in  a hurry to deprecate `calc.lambdat=TRUE` in `reformulas`; removing it would simplify the code a bit, but I don't see any other advantage. (There's also a possibility that some other downstream package is using it, although I doubt it ...)
*

jaganmn:
Just pushed a5d52ca which changed quite a bit (in terms of what is implemented in `R/covariance.R` and how it is organized) so I recommend pulling.  I may not touch things for a while.

`cs` and `ar1` should work now in addition to `diag` (all without `hom=`, until **reformulas** is patched), though I've done minimal testing apart from trying the example at the top of `R/cs_tests.R` in the original branch.  However, see my comment in `TODO.md` about `upper=` and let me know if you have an opinion.

jaganmn:
Is 1042437 specific to this branch?

bbolker:
No, I should cherry-pick it back onto master

bbolker:
Thinking about the `devfun2`/`transfuns` machinery. There are a variety of scales among which we might want to convert:

1\. `par` (optimization-level parameters)
2\.  `theta` (elements of `Lambdat`)
3a\. user-scale, stddev/correlation 
3b\. user-scale, variance/covariance

Right now we're defining a `getTheta` function (`par` → `theta`) for each covariance subclass. For profiling etc., we need to do optimization on one of the two user scales (3a, 3b).  That means we need to first convert the MLE `par` to {3a, 3b} (to get the starting point for optimization), then within the objective function we need to convert {3a,3b} either to theta (which is used in the innermost deviance function calculation), or more likely to `par` (which we'll use in an existing deviance function, which will internally call `getTheta` ...)

So we need at least one function per class that converts either 3a or 3b to `par`. If we don't mind chaining functions we could pick one (e.g. 3a, stddev/correlation) as being canonical, since converting from stddev/correlation to variance/covariance or vice versa is easy and universal.

One other possibly tricky part is that depending on the covariance type there may be different numbers of user-level parameters to profile over -- e.g. in a posdef/unstructured case we want to profile over all (n) std devs and (n*(n-1)/2) correlations, while in the homogeneous AR1 case we only want to profile over (1) stddev and (1) correlation ...

Thoughts?

bbolker:
does it make sense to allow `nc<=0` ?? shouldn't we throw an error in this case (i.e., it most likely indicates something wrong upstream) rather than silently doing something sort of sensible?

jaganmn:
I don't think that the class definition needs to change.  If `reformulas::mkReTrms` expects positive `nc <- lengths(cnms)`, then *it* should test that and conditionally signal an error with some message more useful than `"'nc' is not positive"`.

bbolker:
FYI I'm chipping away at the `devfun2`/profiling stuff in `flexSigmaMinimum_devfun2` (I decided to branch off since my code is pretty ugly right now). First pass, trying to make unstructured covariance stuff work with machinery that should be generalizable ... once that's working I will go back and try to actually make it general (e.g. define the relevant methods for other covariance structure classes)

jaganmn:
I've just added support for `glmer`, which will need its own tests.  Moving on to `nlmer` ...

annahuynhly:
@bbolker I'm writing integration tests by comparing `glmmTMB` to `lme4` and I remember in our conversation that a relative mean difference of a tolerance smaller than `10e-6` is okay. Here, the mean relative differences are only true for a tolerance of `10e-4`; see below:

```{r}
library(glmmTMB)
library(lme4)

fm1 <- lmer(Reaction ~ Days + (Days | Subject), sleepstudy, REML = FALSE)
fm1.glmmTMB <- glmmTMB(Reaction ~ Days + (Days | Subject), sleepstudy, REML = FALSE)

all.equal(c(as.matrix(VarCorr(fm1)[[1]])),
          c(VarCorr(fm1.glmmTMB)$cond[[1]]))
# [1] "Mean relative difference: 2.728569e-05"

# Technically a redundant case below, but want us to act properly
fm2 <- lmer(Reaction ~ Days + us(Days | Subject), sleepstudy, REML = FALSE)
fm2.glmmTMB <- glmmTMB(Reaction ~ Days + us(Days | Subject), sleepstudy, REML = FALSE)
# [1] "Mean relative difference: 2.728569e-05"

all.equal(c(as.matrix(VarCorr(fm2)[[1]])),
          c(VarCorr(fm2.glmmTMB)$cond[[1]]))

fm3 <- lmer(Reaction ~ Days + cs(Days | Subject), sleepstudy, REML = FALSE)
fm3.glmmTMB <- glmmTMB(Reaction ~ Days + cs(Days | Subject), sleepstudy, REML = FALSE)

all.equal(c(as.matrix(VarCorr(fm3)[[1]])),
          c(VarCorr(fm3.glmmTMB)$cond[[1]]))
# "Mean relative difference: 3.610441e-05"
```

Is this okay or does this highlight a problem for `lme4`?

tests/testthat/test-covariance_structures.r
```diff
@@ -1,7 +1,7 @@
 
 test_that("unit test for unstructured covariances", {
   
-  for(i in seq_len(100L)){
+  for(i in seq_len(20L)){
```
jaganmn: On second thought, it will be better to do `for (nc in c(0:4, 16L, 64L, 256L))`, so that you are sure to test the important edge cases and also some big ones.

bbolker:
You should put it in `inst/testdata` (see e.g. `inst/testdata/lme-tst-fits.R`, which has a similar function ...) (*don't* make a separate `flexSigmaMinimum` subdirectory, this stuff will eventually be part of the main package ...)

jaganmn:
You might also be running r-lib/waldo#211.  You will want to reverse the arguments of one of `base::all.equal` and `testthat::expect_equal`, if you are trying to verify that they calculate the same relative error.

annahuynhly:
> You might be running into #861. I'm not sure but maybe you can work around it by passing
> 
> ```
> control = lmerControl(check.nobs.vs.nRE = "ignore"))
> ```

Unfortunately, still doesn't seem to work:
```{r}
fm4 <- lmer(Reaction ~ Daysf + ar1(Daysf+0 | Subject), sleepstudy, 
            REML = FALSE, control = lmerControl(check.nobs.vs.nRE = "ignore"))
```
Produces the following warning message:
```{r}
Warning message:
In sqrt(1 - rho2) : NaNs produced
```
As we can see, this is an obvious issue since:
```{r}
> VarCorr(fm4)
 Groups   Name   Std.Dev. Corr                                                 
 Subject  Daysf0 NaN                                                           
          Daysf1 NaN        NaN                                                
          Daysf2 NaN        NaN   NaN                                          
          Daysf3 NaN        NaN   NaN   NaN                                    
          Daysf4 NaN        NaN   NaN   NaN   NaN                              
          Daysf5 NaN        NaN   NaN   NaN   NaN   NaN                        
          Daysf6 NaN        NaN   NaN   NaN   NaN   NaN   NaN                  
          Daysf7 NaN        NaN   NaN   NaN   NaN   NaN   NaN   NaN            
          Daysf8 NaN        NaN   NaN   NaN   NaN   NaN   NaN   NaN   NaN      
          Daysf9 NaN        NaN   NaN   NaN   NaN   NaN   NaN   NaN   NaN   NaN
 Residual        NaN   
```

annahuynhly:
> How did you install the package? I see:
> 
> ```
> > library(lme4)
> > sleepstudy$Daysf <- factor(sleepstudy$Days)
> > lmer(Reaction ~ Daysf + ar1(0 + Daysf | Subject), sleepstudy,
> +      REML = FALSE, control = lmerControl(check.nobs.vs.nRE = "ignore"))
> Linear mixed model fit by maximum likelihood  ['lmerMod']
> Formula: Reaction ~ Daysf + ar1(0 + Daysf | Subject)
>    Data: sleepstudy
>       AIC       BIC    logLik -2*log(L)  df.resid 
> 1831.4714 2042.2066 -849.7357 1699.4714       114 
> Random effects:
>  Groups   Name   Std.Dev. Corr                                        
>  Subject  Daysf0 33.229                                               
>           Daysf1 34.031   0.81                                        
>           Daysf2 27.412   0.66 0.81                                   
>           Daysf3 33.534   0.53 0.66 0.81                              
>           Daysf4 37.308   0.43 0.53 0.66 0.81                         
>           Daysf5 50.235   0.35 0.43 0.53 0.66 0.81                    
>           Daysf6 67.576   0.29 0.35 0.43 0.53 0.66 0.81               
>           Daysf7 53.287   0.23 0.29 0.35 0.43 0.53 0.66 0.81          
>           Daysf8 58.614   0.19 0.23 0.29 0.35 0.43 0.53 0.66 0.81     
>           Daysf9 62.426   0.15 0.19 0.23 0.29 0.35 0.43 0.53 0.66 0.81
>  Residual         2.245                                               
> Number of obs: 180, groups:  Subject, 18
> Fixed Effects:
> (Intercept)       Daysf1       Daysf2       Daysf3       Daysf4       Daysf5  
>     256.652        7.844        8.710       26.340       31.998       51.867  
>      Daysf6       Daysf7       Daysf8       Daysf9  
>      55.526       62.099       79.978       94.199  
> ```

I installed it using
```{r}
R CMD build .
R CMD INSTALL lme4_1.1-37.9000.tar.gz
R
```
I still get the same issue...

> I also see:
> 
> ```
> > library(lme4)
> > library(nlme)
> > fm4 <- lmer(Reaction ~ Days + cs(Days | Subject), sleepstudy)
> > fm4.nlme <- lme(fixed = Reaction ~ Days, random = ~Days | Subject,
> +                 correlation = corCompSymm(form = ~Days | Subject),
> +                 data = sleepstudy)
> > (x <- c(as.matrix(VarCorr(fm4)[[1]])))
> [1] 612.160221   9.613537   9.613537  35.070442
> > (y <- c(getVarCov(fm4.nlme)))
> [1] 612.079511   9.606036   9.606036  35.071302
> > all.equal(x, y)
> [1] "Mean relative difference: 0.0001449029"
> > all.equal(y, x)
> [1] "Mean relative difference: 0.0001449235"
> ```
The main issue here isn't the value (I get the same ones); it's that Ben suggested the tolerance of their differences should be `5e-5` in which this does not fall under.

annahuynhly:
Another thing to report: results from `glmmTMB` and `lme4` with variance-covariance structure `ar1` seem to report drastically different values; edited comment here: https://github.com/lme4/lme4/pull/877#issuecomment-3503043483 since I don't want to lose track of those issues... (Have my own comment bookmarked for my next meeting :'))

jaganmn:
> Another thing to report: results from `glmmTMB` and `lme4` with variance-covariance structure `ar1` seem to report drastically different values

In **glmmTMB**, `ar1` and `hetar1` in a formula specify homogeneous and heterogeneous variance, respectively; see, e.g., `help("glmmTMB")`.  Does that help?

bbolker:
* easy enough to change the names

The code that translates optimization-scale parameters to profile-scale parameters is currently using a scheme that goes ( optimization pars → SD/cor scale → optimization pars → SD/cor scale → profile-scale parameters), using `setProfPars`/`setVC`/`getVC`/`getProfPars`, and furthermore does it by modifying the fitted object and then extracting the new parameters from the modified object. At a quick glance it is giving reasonable results (haven't checked anything carefully), but it feels very clunky, suggesting that the machinery could be refactored/redesigned to make things smoother.
 
This is the code to get from `ores$par` (optimization scale) to `ppars` (profiling scale)

```r
 setProfPars(fitted, c(ores$par, sig), profscale = prof.scale, sc = sig)
 ppars <- unlist(getProfPars(fitted, profscale = prof.scale, sc = sig))
```

tests/testthat/test-covariance_structures.r
```diff
@@ -361,13 +393,80 @@ test_that("integration tests for sigma", {
...
+  ## we see that the mean relative difference between fitted models from 
+  ## different packages will be quite large.
+  ## Idea: if log likelihoods are the same, then we're okay.
+  expect_equal(logLik(fm1), logLik(fm1.us))
+  expect_equal(logLik(fm1.REML), logLik(fm1.us.REML))
+  expect_equal(logLik(gm), logLik(gm.us))
+  
+  ## comparing against glmmTMB
+  expect_true(all.equal(as.numeric(logLik(fm1)), other_mod$fm1.glmmTMB_logLik, 
+                        check.class = FALSE,
```
jaganmn: Since you use `all.equal(..., check.attributes = FALSE, check.class = FALSE)` in many places, maybe it is worth defining and using a wrapper function, something like:

```r
all.equal.ca0 <-
function (...)
all.equal(..., check.attributes = FALSE, check.class = FALSE)
```

annahuynhly: Good call; would make the code a lot cleaner. Will soon.

bbolker:
Re https://github.com/lme4/lme4/pull/877#issuecomment-3520480583: thanks, I'll try to integrate this into `R/profile.R` as soon as I can ...

annahuynhly:
Edit: actually, might not be a mistake...  [6a56823](https://github.com/lme4/lme4/pull/877/commits/6a56823c03044a2f08c1f20b15c114ddfe6d8399), basically I tried to push something before pulling (on accident). I think that just did a merge with what I tried doing...

annahuynhly:
I'll look into it... I believe errors like:
```{r}
all.eqNum(vcov(fm1), other_mod$fm1.glmmTMB_vcov) not equal to 0.
```
Signify that instead the output is TRUE instead of a number, signalling that they're exactly equal. I'll think about ways to deal with it tomorrow...

bbolker:
now passes at test level 1, but fails at test level 2 (more profiling stuff, sigh).

bbolker:
Recording this for now since my brain is fried. 

I tried going back from `safe_chol` to `chol` in `setVC()` to see what was happening. At the best I can tell at the moment it looks (???) like the optimizer may be slightly violating its allowed bounds, i.e. trying a correlation parameter of 1.0001. (I have known this to happen when using finite difference approximations for the gradient but it surprises me here ...)

Anyway, will work on this when I get a chance ...

NAMESPACE
```diff
@@ -156,11 +160,16 @@ export(.prt.VC,		.prt.aictab,	.prt.call,	.prt.family,
 ##------ Our S4 Classes ------------------------
 exportClasses(glmerMod, lmerMod, nlmerMod, merMod)
 exportClasses(lmList4)
+exportClasses(Covariance, Covariance.us, Covariance.diag, Covariance.cs,
```
jaganmn: I'm fine with exporting the classes, but doing so requires that they are documented.  You need to create a `man/Covariance-class.Rd` containing a minimal amount of information, something like (filling in the `TODO`):

```
\name{Covariance-class}
\alias{Covariance-class}
\alias{Covariance.us-class}
\alias{Covariance.diag-class}
\alias{Covariance.cs-class}
\alias{Covariance.ar1-class}
\docType{class}
\title{Virtual Class \sQuote{Covariance} of Covariance Matrices}
\description{TODO.}
\section{Slots}{
  \describe{
    \item{\code{nc}}{TODO.}
    \item{\code{par}}{TODO.}
    \item{\code{hom}}{TODO.}
  }
}
\keyword{classes}
\keyword{internal}
```

If we wanted to document (mainly for ourselves) what each of the generic functions does with objects of these classes, then we could add:

```
\section{Methods}{
  \describe{
    \item{\code{getPar}}{TODO.}
    %% and so on
  }
}
```

NAMESPACE
```diff
@@ -144,6 +144,10 @@ export(updateGlmerDevfun)
 export(VarCorr)
 export(getData)
 export(varianceProf)
+export(getTheta, getThetaLength, getThetaIndex, getThetaIndexLength,
```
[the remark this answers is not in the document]
annahuynhly: Sure; I'll undo the changes (if not already).

NAMESPACE
jaganmn: Rip

bbolker:
passing `R CMD check` at `LME4_TEST_LEVEL`=10 now (!!)  (although I am still getting a warning about undocumented classes ...)

A small request for the `chol()`/`setPar()` machinery: since `lme4` is all set up around being able to compute log-likelihoods etc. properly on the boundary, it would be nice if setting correlation == 1 gave a non-NA result (I have added a `try()` clause to the results of `devfun2()` so that it returns NA if the parameter conversion fails.

Example:

```r
m1 <- lmer(Reaction ~ Days + (Days | Subject), sleepstudy)
dd <- devfun2(m1)
par1 <- c(16.066, 4.744, 0.955, 27.525)
mkpar <- function(rho) { par1[3] <- rho; par1 }
dd(par1)
dd(mkpar(0.999))
dd(mkpar(1.0)) ## NA
```

As for my [comment above about `upReCovs`/back-compatibility](https://github.com/lme4/lme4/pull/877#issuecomment-3538748032); I'm not 100% sure if the problem is actually replicable &mdash; I will try to be a bit more systematic, if it goes away I won't worry about it.

annahuynhly:
Some further `ar1` exploration; I thought the problem earlier was that `sleepstudy` was a faulty dataset to assume an `ar1` covariance structure. However, with a second example I'm thinking I likely do not have the `nlme` equivalent (or maybe both `glmmTMB` and `lme4` have a similar error.)

```{r}
library(lme4)
library(glmmTMB)
library(nlme)

data("Soybean", package = "MEMSS")

Soybean$Timef <- factor(Soybean$Time, ordered = TRUE)

soy.lme4 <- lmer(weight ~ Timef + ar1(0 + Timef | Plot, hom = TRUE), 
            Soybean, REML = FALSE, 
            control = lmerControl(check.nobs.vs.nRE = "ignore"))

soy.glmmTMB <- glmmTMB(weight ~ Timef + ar1(0 + Timef | Plot), 
                       Soybean, REML = FALSE)

soy.nlme <- gls(weight ~ Timef,
                correlation = corExp(form = ~ Time | Plot, nugget = TRUE),
                data = Soybean, method = "ML")

## Comparing lme4 to glmmTMB ########################
# likelihoods are similar; below returns TRUE
all.equal(-soy.glmmTMB$fit$objective, logLik(soy.lme4), 
          check.attributes = FALSE,
          check.class = FALSE)

all.equal(vcov(soy.lme4), vcov(soy.glmmTMB)$cond, check.attributes = FALSE,
          check.class = FALSE)
# "Mean relative difference: 0.0001230642"
# not 5e-5, but should be ok?

## Comparing lme4 to nlme ###########################
# differences are relatively large...
all.equal(logLik(soy.lme4), logLik(soy.nlme), check.attributes = FALSE,
          check.class = FALSE)
# [1] "Mean relative difference: 0.012905"
all.equal(vcov(soy.lme4), vcov(soy.nlme), check.attributes = FALSE,
          check.class = FALSE)
# "Mean relative difference: 0.3676962"

#################################
## second version (try using REML!!!)

soy.lme4.REML <- lmer(weight ~ Timef + ar1(0 + Timef | Plot, hom = TRUE), 
            Soybean, control = lmerControl(check.nobs.vs.nRE = "ignore"))

soy.nlme.REML <- gls(weight ~ Timef,
                correlation = corExp(form = ~ Time | Plot, nugget = TRUE),
                data = Soybean)
# doesn't improve...
all.equal(logLik(soy.lme4.REML), logLik(soy.nlme.REML), 
          check.attributes = FALSE, check.class = FALSE)
# [1] "Mean relative difference: 0.01200982"
all.equal(vcov(soy.lme4.REML), vcov(soy.nlme.REML), 
          check.attributes = FALSE, check.class = FALSE)
# [1] "Mean relative difference: 0.2913537"
```

---

Another thing to note is that the results from `nlme` and `lme4` differ quite a bit when it comes to nonlinear mixed effects models. I may again not have the proper equivalent. 

Consider the following `lme4` models:
```{r}
startvec <- c(Asym = 200, xmid = 725, scal = 350)
nm <-  nlmer(circumference ~ SSlogis(age, Asym, xmid, scal) ~ Asym|Tree,
              Orange, start = startvec)

nm.us <- nlmer(circumference ~ SSlogis(age, Asym, xmid, scal) ~ us(Asym|Tree),
               Orange, start = startvec)

nm.cs <- nlmer(circumference ~ SSlogis(age, Asym, xmid, scal) ~ cs(Asym|Tree),
               Orange, start = startvec)
```

And the following `nlme` equivalents:
```{r}
startvec <- c(Asym = 200, xmid = 725, scal = 350)
nm.nlme <- nlme(
  circumference ~ SSlogis(age, Asym, xmid, scal),
  data = Orange,
  fixed = Asym + xmid + scal ~ 1,   
  random = Asym ~ 1 | Tree, 
  start = startvec)

nm.nlme.cs <- nlme(
  circumference ~ SSlogis(age, Asym, xmid, scal),
  data = Orange,
  fixed = Asym + xmid + scal ~ 1,   
  random = Asym ~ 1 | Tree, 
  correlation = corCompSymm(form = Asym ~ 1 | Tree),
  start = startvec)
```

R/profile.R
```diff
@@ -481,11 +484,18 @@ get.which <- function(which, nvp, nptot, parnames, verbose=FALSE) {
...
-                    transfuns = list(from.chol = Cv_to_Sv,
-                                       to.chol = Sv_to_Cv, to.sd = identity),
+devfun2 <- function(fm,
+                    useSc = if(isLMM(fm)) TRUE else NA,
+                    scale = c("sdcor", "varcov"),
                     ...)
 {
+
+    scale <- match.arg(scale)
+    if (scale == "varcov" && anyStructured(object = fm)) {
```
jaganmn: FWIW, methods for `Covariance.diag` do support `profscale="varcov"`.

R/profile.R
```diff
@@ -514,17 +521,16 @@ devfun2 <- function(fm, useSc = if(isLMM(fm)) TRUE else NA,
...
-            sigsq <- sigma^2
-            pp$ldL2() - ldW + (resp$wrss() + pp$sqrL(1))/sigsq + n * log(2 * pi * sigsq)
+      {
+          ## convert 'profPar' -> 'pars'
+          sig <- pars[length(pars)]
+          thpars <- try(
+              convProfParToPar(pars, fm, profscale = scale, sc = sig),
+              silent = TRUE)
+          if (inherits(thpars, "try-error")) return(NA)
+          .Call(lmer_Deviance, pp$ptr(), resp$ptr(), thpars)
```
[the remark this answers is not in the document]
jaganmn: @bbolker This might explain the discrepancy that you observed in https://github.com/lme4/lme4/pull/877#issuecomment-3555477872.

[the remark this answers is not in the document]
jaganmn: Fixed with e8c9c43.

bbolker:
Last commit helps (we get past the base-deviance check) but results are still wonky, will continue to dig ...

annahuynhly:
> > I was looking at the methods for class `merMod` and I feel like if someone is going to use `getPar` and `getTheta` on a `merMod` object, they might want to use `getLambda` too.
> 
> The need hasn't arisen, but if we did have such a method then it should return the transpose of the `dgCMatrix` object constructed by `upReCovs`, which is a bit more involved than your proposal but probably not too hard.

I feel like I am not following; `upReCovs` is:

```{r}
upReCovs <-
function (reCovs, theta) {
    ## FIXME?  function (reCovs, par) instead?
    pos <- 0L
    for (i in seq_along(reCovs)) {
        elt <- reCovs[[i]]
        reCovs[[i]] <- setTheta(elt, theta, pos)
        pos <- pos + getThetaLength(elt)
    }
    reCovs
}
```

And thus both would give the same output (I'm not seeing the `dgCMatrix` object):

```{r}
library(lme4)
fm1.us <- lmer(Reaction ~ Days + us(Days | Subject), sleepstudy)
getReCovs(fm1.us)
upReCovs(getReCovs(fm1.us), getTheta(fm1.us))
```

jaganmn:
I wouldn't mind merging into master now, in order to force our hand a bit.  If we want to continue to discuss things in one place then we can create a TODO issue.

PR #892, opened 2025-12-06
revert double vert default to "split"

bbolker:
I don't think this breaks anything, and it should un-break the downstream packages that are assuming that double-bar means "split into separate terms" (`(1+a+b||f)` → `(1|f) + (0 + a|f) + (0 + b|f)`). Still needs to be documented better ... i.e. that default behaviour can be set by `options()` and retrieved by `getDoublevertDefault` ...

### DIFF
--- a/tests/testthat/test-doubleVertNotation.R
+++ b/tests/testthat/test-doubleVertNotation.R
@@ -6,15 +6,15 @@ m1 <- lmer(form1, sleepstudy)
 m2 <- lmer(form2, sleepstudy)
 
 test_that("basic intercept + slope '||' works", {
-    if (FALSE) {
-    ## '||' is no longer expanded as two '|'
-    expect_equivalent(lFormula(form1, sleepstudy)$reTrms,
-                      lFormula(form2, sleepstudy)$reTrms)
-    }
-    expect_equivalent(fitted(m1),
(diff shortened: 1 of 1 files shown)


jaganmn:
I'm ultimately not too opinionated about `||`, as long as everything is eventually clearly documented, so feel free to merge.  I'll just add that **lme4**-using packages should be encouraged to write code like

```
lmerWrap <-
function (...) {
    oo <- options(lme4.doublevert.default = "diag")
    on.exit(options(oo), add = TRUE)
    lmer(...)
}
```

and not do

```
.onLoad <-
function (libname, pkgname) {
    options(lme4.doublevert.default = "diag")
    NULL
}
```

which could disrupt other **lme4**-using packages and actually create reproducibility headaches.  Having said that, did you consider adding a way for the caller to override any global option with an argument to the call, as with `print(x, digits=*)`?

jaganmn:
I'm happy to wait and see how this release goes and to delay adding complexity until we know that it's needed.

> The tests that were deactivated due to the breaking change could be reactivated.

There was one in `tests/testthat/test-lmer.R` and another in `tests/testthat/test-doubleVertNotation.R`.

bbolker:
tests are restored now.

PR #899, opened 2026-01-03
Bootmer order fix

bbolker:
There may be a little of chaff here due to git confusion but I'm hoping it's OK now.

This also contains (too much) other stuff, including ...

* NEWS and documentation updates
* `confint(., method = "boot")` is a little more robust &mdash; works better when there is zero variance in estimated coefficients from a parametric bootstrap run (as it turns out we don't need this once other problems are fixed, but it shouldn't hurt)
* renamed the covariance structure vignette to make it more general (hopefully the first half of the vignette [not written yet] can become a general description of covariance structures, leaving the second half for the explicit glmmTMB vs lme4 comparison).
* I tried changing the version number from 2.0.0 to 2.0-0 (which is more consistent with our previous naming convention), but then changed my mind and changed it back to 2.0.0 (which I think has become the more general convention in the R world).
* `confint(., method = "wald")` (lowercase) now works (trivial but perhaps useful)
* URL fixes for `--as-cran` testing (mostly http to https)

### DIFF
--- a/tests/testthat/test-covariance_structures.R
+++ b/tests/testthat/test-covariance_structures.R
@@ -412,10 +412,11 @@ test_that("integration tests for coef and fixef", {
   expect_equal(fixef(gm), fixef(gm.us))
   
   ## One of the expected summaries
+  opt <- options(useFancyQuotes = FALSE)
   tmpf <- function(x) capture.output(print(summary(x),digits=1))
   tfun <- function(cc) {
     w <- grep("Fixed effects:", cc)
-    gsub("[‘’]", "'", cc[w:length(cc)])
+    cc[w:length(cc)]
   }
   expected_summary <- c("Fixed effects:",                         
                         "            Estimate Std. Error t value",
@@ -440,6 +441,7 @@ test_that("integration tests for coef and fixef", {
                      "age    -0.022       ",      
                      "urbanY -0.648  0.006")
   expect_equal(tfun(tmpf(gm.us)), expected_sum2)
+  options(opt)
 })
 
 test_that("integration tests for sigma", {


R/lmer.R
```diff
@@ -2120,9 +2120,10 @@ getME.merMod <- function(object,
            "beta"  = object@beta,
            "theta" = setNames(th, getThetaNames(object)),
            "par" = object@optinfo$val[seq_len(getParLength(object))],
```
jaganmn: If names are desired then change to `setNames(., getParNames(object))` for `par` here and change to `setNames(., c(getParNames(object), if (is(object, "glmerMod")) names(fixef(object))))` for `devarg` below.

We've been pretty consistent about using `par` to refer to a vector of just covariance parameters.  It's a bit awkward because people accustomed to `optim` might expect it to refer to the vector of all parameters, but I don't think that we should reverse course now.

It would be nice if `getME` had a `names=FALSE` option, since constructing names is not necessarily cheap, but whatever.

inst/NEWS.Rd
```diff
@@ -4,7 +4,43 @@
...
+      terms, so the behaviour will match \code{afex::mixed} or
+      \code{glmmTMB} (see \code{\link{getDoublevertDefault}}).
+    } % itemize
+  } % new features
+  \subsection{USER-VISIBLE CHANGES}{
+    \itemize{
+      \item code for pretty-printing covariance matrices, especially
+      structured covariance matrices, has moved to the \code{reformulas}
+      package (so functionality can be shared with the \code{glmmTMB} package)
+      \item version number format changed from \code{x.y-z} to \code{x.y.z}
```
jaganmn: FWIW, I like the legacy format because **Matrix** will continue to use it and **lme4** descended from **Matrix**.  *And* we are planning to release **Matrix** 2.0-0 this year ... !!

But I don't mind either way.

R/profile.R
```diff
@@ -1,16 +1,35 @@
...
+  ntp <- length(object@theta)
+  res <- if (signames) {
+           sprintf(".sig%02d", seq(ntp))
+         } else {
+           vv <- getVCNames(object, prf=prefix, old=FALSE)
+           ## result is stored as {vcomp, ccomp}; 'zip' it
+           unlist(
+             lapply(seq_along(vv$vcomp),
+                    function(i) c(vv$vcomp[[i]], vv$ccomp[[i]]))
+           )
```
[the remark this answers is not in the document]
bbolker: I guess this works, but it seems obscure. I prefer what's written here, even though it's a little longer. Maybe `unlist(Map(c, vv$vcomp, vv$ccomp))` is a better substitute?

inst/NEWS.Rd
bbolker: OK, switching back to legacy ...

Issue #886, opened 2025-12-02
Changes to worse in reverse dependencies under lme4 2.0.0


annahuynhly:
The sequel to https://github.com/lme4/lme4/issues/882...

See [CHANGES_1.1-38_2.0.0.tar.gz](https://github.com/user-attachments/files/23869513/CHANGES_1.1-38_2.0.0.tar.gz)

jaganmn:
The number of broken packages is 20 but the number of distinct causes seems small.

PS: Nice work. :-)

jaganmn:
Among the 20 broken strong reverse dependencies I count 9 distinct problems:
|Package|Cause|Location|
|---:|:---|:---|
|**afex**|reflects old `\|\|` behaviour|`tests/testthat/test-mixed-structure.R`|
|**blme**|tests formal arguments of deviance functions and `lme4::updateStart`; does `checkConv(ctrl=NULL)` which is not supported; see also 2c8b480|`blme::blmer`|
|**designr**|reflects old `\|\|` behaviour|`vignette("power-analyses")`, `designr::extractTerms`|
|**GLMpack**|uses `cs` as non-special in model formula|`example("mexico")`|
|**lme4breeding**|reflects old `\|\|` behaviour|`vignette("lmebreed.gxe")`|
|**lmerTest**|tests formal arguments of deviance functions; see also 2c8b480|`lmerTest::as_lmerModLmerTest`|
|**merDeriv**|`nAGQ <= 25L` is newly tested by `getME(., "devfun")`|`inst/tinytest/test_merDeriv.R`|
|**REndo**|tests for `VarCorr` error (which was a bug) that no longer occurs|`tests/testthat/test-inputchecks_multilevel.R`, `REndo::multilevelIV`|
|rstanarm|reflects old `\|\|` behaviour|`tests/testthat/test_stan_functions.R`, local `test_lme4`|

bbolker:
For `GLMpack`: I *might* be able to rework the logic in `reformulas` to ignore specials unless they're at the head of an expression containing `|` ...  (i.e. that would distinguish between `cs(f)` and `cs(1|f)`)

jaganmn:
> We could improve back-compatibility for the old `||` behaviour by allowing an option setting that would specify the `expand_doublevert_method` argument to `findbars_x` with `lFormula`/`glFormula` ...

That makes sense.  For the maintainers, enabling the option will be much easier than adapting to the new behaviour.

[...]

Issue #852, opened 2025-07-29
modify ?print.VarCorr.merMod for GLMMs with scale parameter?


bbolker:
this function prints out 'dispersion estimates' as "Residual std dev" whether they're really Residual std dev (i.e. LMMs) or just `sigma()` output (e.g. Gamma GLMMs) ...

annahuynhly:
I realised upon testing that `VarCorr()` is for `lme4` is incorrect when using `glmer` (ugh) by comparing the output to `glmmTMB`.

```{r}
library(lme4)
library(glmmTMB)

# borrowed this function from test-glmFamily
simfun_gam <- function(ngrp = 50, nrep = 50, shape_gam = 2, intercept = 1, 
                       theta_val = 2, use_simulate = FALSE, seed = NULL) {
  if (!is.null(seed)) set.seed(seed)
  dd <- expand.grid(group = 1:ngrp, rep = 1:nrep)
  if (use_simulate) {
    dd$y <- simulate(~ 1 + (1 | group), newdata = dd, 
                     family = Gamma(link = "log"), 
                     newparams = list(
                       theta = theta_val, beta = 1, sigma = 1/sqrt(shape_gam)))[[1]]
    return(dd) 
  }
  b <- rnorm(ngrp)
  eta <- intercept + b  
  mu <- exp(eta)
  y <- rgamma(nrow(dd), shape = shape_gam, scale = mu/shape_gam)
  data.frame(dd, y)
}

dd2 <- simfun_gam(seed = 101, use_simulate = TRUE)

m1 <- glmer(y ~ 1 + (1|group), family = Gamma(link = "log"), data = dd2)
VarCorr(m1)
m1@reCovs[[1]]@par

g1 <- glmmTMB(y ~ 1 + (1|group), family = Gamma(link = "log"), data = dd2)
VarCorr(g1)
```
In the above example, ideally for both values we should have an output of 2.

I'll be working on this to make `VarCorr()` printing more consistent between `glmmTMB` and `lme4`.

Also, there doesn't seem to be an issue for `lmer`.

PR #904, opened 2026-01-11
Editing the Covariance Vignette

annahuynhly:
According to: https://github.com/lme4/lme4/issues/886#issuecomment-3708426807 

I gave more details in the vignette; hopefully this is what you had in mind?

Unsure if I needed to give actual technical details on how the code works; but I feel like at that point, someone should just peek at `R/covariance.R`.

### DIFF
--- a/vignettes/covariance_structures.Rmd
+++ b/vignettes/covariance_structures.Rmd
@@ -1,3 +1,4 @@
+
 ---
 title: "Covariance Structures"
 output: rmarkdown::html_vignette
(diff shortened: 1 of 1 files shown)


annahuynhly:
I realize that I likely should add a section detailing that the construction
```math
\mathbf{\Sigma}_{\mathbf{\theta}} = \sigma^{2} \Lambda_{\theta} \Lambda_{\theta}^{T}
```
is not applicable to GLMMs. I'll add this when I fix the bug we discussed here https://github.com/lme4/lme4/issues/852

Issue #928, opened 2026-03-01
WISH: Add support for future-based parallelization


HenrikBengtsson:
@bbolker wrote in https://github.com/lme4/lme4/issues/924#issuecomment-3980961433:

> Longer term, it would be nice to `future`-enable our parallel machinery ... 

That's exciting to hear - I'm happy to provide a PR. To add support for the future ecosystem (aka "futureverse"), the easiest is to rely on **future.apply**, e.g.

```r
               future.apply::future_lapply(seq_fit, ffun)
```

(*) We can make it even easier using the new **futurize** package. See below.

In the API, I can imagine the easiest would be to add support for `parallel = "future"`. What do you think?

I see that `future` is relatively light in terms of _strong_ dependencies ...
> 
> > length(package_dependencies("future", recursive=TRUE)[[1]])
> [1] 8
> > length(package_dependencies("future", recursive=TRUE, which = "most")[[1]])
> [1] 2300

Yes, depending on **future.apply** would bring in the following seven dependencies for the end user:

```
[1] "parallelly"   "future.apply" "listenv"      "codetools"   
[5] "digest"       "globals"      "future"      
```

UPDATE +30min: Dependency **digest** is actually not used in R (>= 4.5.0), where `tools::md5sum()` is used. I plan to move **digest** to Suggests: when R 4.7.0 is released in 2027.  Dependency **codetools** ("recommended") comes from **globals** - it might become an optional dependency in a future release of **globals**, but that is to be discussed. Finally, dependency **listenv** is not needed for most future-based code  - it might become an optional dependency ("Suggests:") in the future, but that's to be discussed.

For **lme4**, the only dependency is **future.apply** itself, either under `Imports:` or `Suggests:` (your call).

---

Regarding **futurize** comment above.  To lower the mental load and minimize code clutter, the absolutely simplest code update would be to use:

```r
               lapply(seq_fit, ffun) |> futurize::futurize()
```

That keeps the original logic as-is. This transpiles into the requires `future.apply::future_lapply()` call automatically. Taking this approach would add dependencies **futurize** and **future.apply**.

You can punt on **futurize** for now.

HenrikBengtsson:
I looked into what it would take to add support for `parallel = "future"`. In that process, I spotted a few things that could/should be improved in the current code. I've created backward compatible PR #929 for this. FWIW, if merged, adding `parallel = "future"` would be a walk in the park.

annahuynhly:
[...]

PR #929, opened 2026-03-02
Clean up and robustify parallelization code

HenrikBengtsson:
* Drop use of `do_parallel`, `have_mc` and `have_snow`; rely on `parallel` instead

* Validate also arguments `cl` and `ncpus`

* Use `on.exit(parallel::stopCluster(cl))` everywhere. This avoids leaving stray clusters behind in case of run-time errors

This also prepares for future additions, e.g. `parallel = "future"` (#928)

### DIFF
--- a/R/allFit.R
+++ b/R/allFit.R
@@ -113,9 +113,9 @@ allFit <- function(object, meth.tab = NULL,
 	      is.character(method    <- meth.tab[,"method"]))
 
     parallel <- match.arg(parallel)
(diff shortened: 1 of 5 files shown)


bbolker:
We're running revdep checks (@jagamn has mostly been handling it for me these days). I've tried a variety of strategies (`revdepcheck` package, other hand-rolled alternatives, `tools::check_packages_in_dir()`) ... no matter what I do there seem to be a few packages that slip between the cracks, although we do catch most of the problems before we submit to CRAN ...

PR #931, opened 2026-03-02
ROBUSTNESS: parallel::detectCores() might return 1L or NA_integer_

HenrikBengtsson:
UPDATE +30min: I've also removed a dependency on the **snow** package, which stemmed from the use of `type = "SOCK"` in `optCls <- makeCluster(nc, type = "SOCK")` part of one of the examples.


bbolker:
I'm going to take a chance ...

Issue #873, opened 2025-10-13
lFormula() and glFormula() clean-up


annahuynhly:
The following comments in the code: https://github.com/lme4/lme4/blob/905748659cf4010f92b0708bd5c42fc9c366bf6e/R/modular.R#L637C1-L638C88 suggest that we should merge `lFormula()` and `glFormula()`. Also, should remove unnecessary arguments.

Issue #832, opened 2025-06-05
switch to roxygen for documentation?


bbolker:
(pinging @mmaechler who will have the strongest feelings about this ...)

I know some of us have always preferred `.Rd`-format documentation; I would like to consider switching to roxygen (which can be semi-automated by using the [Rd2roxygen package](https://cran.r-project.org/package=Rd2roxygen); there is also someone who is willing to take on the task).

The main advantage of using roxygen, from my point of view, is that it makes it easier to maintain consistency between the code and the docs. The IDE support (ESS/RStudio/etc.) for editing roxygen in-place is pretty good now ...

mmaechler:
Yes, I would really be unhappy (and frustrated a bit) about this:
The effect of using Roxygen instead of Rd  often are much poorer help pages:
Using lists (itemize / describe) and nesting them .. using math equations,  careful markup for references, all things I have *not* seen in Roxygen generated help pages ((even though I'm sure they exist too)).
I see Roxygen as an easy way for programmers who don't really like to spend time for writing good (for humans!) help files.

mmaechler:
because it is very awkward to write  multiline \describe{  \item  ... \item ... }  all in *comments* of R code which is how Roxygen has to be done.    
This is the reason all (really all?) roxygen based help pages look so much like written by an AI which cannot know what it's talking about (yes, this is slightly exaggerated .. ;-))

jaganmn:
I don't have a real horse in this race, not being an author, but since I've been offered a soapbox ...

I generally prefer Rd in part for the reason that Martin mentioned.  Formatting LaTeX-like code in a comment block is quite awkward because (at least) ESS does not know how to correctly match braces across comment lines.  But there is another reason not yet mentioned.

Being accustomed to doing `svn log -v` or `git log --name-status`, I like being able to know from just that output, at a glance, whether a commit affected code, documentation, or both.  Keeping code and documentation in independently modified files allows for that.  Keeping them in dependently modified files generally requires that I inspect the output of `svn diff` or `git diff` "to be sure".

On top of that, when changes in documentation appear twice in a diff (because `.Rd` is compiled from `.R` and both are under version control), I am often compelled to suppress changes to `.Rd` from diff output, finding them redundant and distracting.  (And of course there is no such control when one reviews a PR through GitHub.)

Issue #81, opened 2013-08-14
remove unnecessary Roxygen material


bbolker:
As of 13-Aug-2013 we have decided to make `.Rd` files the primary documentation; only basic Roxygen material (parameter definitions?) will be kept in the R code.  At some point we should go in and strip this stuff out.

PR #954, opened 2026-03-20
lFormula() and glFormula() clean-up...

annahuynhly:
In an issue I realized in some old comments that some features from `lFormula` and `glFormula` are similar and ideally should be merged (see: https://github.com/lme4/lme4/issues/873)

Tested with copilot in my fork here: https://github.com/annahuynhly/lme4/pull/4 

With various tests (i.e., `devtools::test()` and directly running `R CMD check` on a tarball, it seems like this code reasonably works and does reduce a lot of redundant code by making a common helper function `mkFormula` for both `lFormula` and `gFormula`.


R/modular.R
bbolker: I'm actually not sure we want this fix.  Because `lme4` doesn't use roxygen (see https://github.com/lme4/lme4/issues/81, https://github.com/lme4/lme4/issues/832), having 'proper' roxygen formatting actually runs the risk of accidentally overwriting the `?modular` documentation at some point (I don't know if copilot was smart enough to figure this out &mdash; it's an unusual case in the universe of R code)

bbolker: (comment was meant to refer to the "fix roxygen formatting" commit,  not the rest of the changes ...)

bbolker: i.e. [this code](https://github.com/lme4/lme4/pull/954/changes/e4e35cd27ce9ed789c19b574f167fdca5b7c15f5#diff-6102084076fc4e633832435611811438b40b4da2ee6205b101aab0047ee26761R489-R491) is also repeated between `[g]lFormula`, I think, but involves `parent.frame()` ... I think `parent.frame` "chains" properly, but should test (coming up with appropriate test cases for scoping/environment-related stuff is tricky ...)

bbolker:
I can imagine I might have to augment the weights-setting machinery [here](https://github.com/lme4/lme4/compare/master...lFormula_env#diff-1133b4ed618525dc31392ae5cf9ca97832d30b2a926d2c4d007e1d3166d56fa4R709-R711) with something analogous for offsets ...

bbolker:
```r
old <- list()
for (var in c("weights", "offset")) {
    old[[var]] <- if (exists(var, environment(formula), inherits = FALSE)) {
        get(var, environment(formula), inherits = FALSE)
    } else {
        list(NULL)  ## note https://cran.r-project.org/doc/FAQ/R-FAQ.html#How-can-I-set-components-of-a-list-to-NULL_003f
    }
}
on.exit(
    Map(function(name, value) {
         if (is.null(value)) {
              rm(list = name, envir = environment(formula))
        } else {
            assign(name, value, envir = environment(formula))
        }
     },  names(old), old),
 add = TRUE
    )
## OR ...
on.exit(
    lapply(names(old),
      function(name) {
         value <- old[[name]]
         if (is.null(value[[1]])) {
              rm(list = name, envir = environment(formula))
        } else {
            assign(name, value, envir = environment(formula))
        }
     }),
 add = TRUE
    )
```

annahuynhly:
@bbolker I think it's good to review now.

I don't know why it says it also removed `glmer_boot_sigTRUE.RData`; I don't think I ever committed...?

bbolker:
that file must have gotten committed by accident at same point (don't know by whom).

Can you try 

```
R CMD build --compact-vignettes=both lme4; export LME4_TEST_LEVEL=20; R CMD check lme4_2.0-2.tar.gz
```
(after merging again)?
I think we would have caught some of the stuff I've been dealing with this morning if we had tested at the higher test level ...

R/modular.R
```diff
@@ -399,68 +377,121 @@ lFormula <- function(formula, data=NULL, REML = TRUE,
...
+    dots <- list(...)
+    do.call(checkArgs, c(list("lmer"), dots[!names(dots) %in% dontChk]))
+    if (!is.null(dots[["family"]])) {
+        ## lmer(...,family=...); warning issued within checkArgs
+        mc[[1]] <- quote(lme4::glFormula)
+        if (missing(control)) mc[["control"]] <- glmerControl()
+        return(eval(mc, parent.frame()))
+    }
+
+    cstr <- "check.formula.LHS"
```
bbolker: is there a reason lines 472-484 can't be moved inside mkFormula?  How does this part of the processing differ between LMMs and GLMMs?  (The guiding principle is that the only parts *outside* of mkFormula should be LMM/GLMM-specific pieces ...)

R/modular.R
```diff
@@ -737,110 +764,36 @@ glFormula <- function(formula, data=NULL, family = gaussian,
...
                              checkLHS = control$check.formula.LHS == "stop")
     formula <- as.formula(formula, env = denv) # substitute evaluated version
-    if (getDoublevertDefault() == "split") {
-      RHSForm(formula) <- reformulas::expandDoubleVerts(RHSForm(formula))
-    }
+    if (getDoublevertDefault() == "split")
+        RHSForm(formula) <- reformulas::expandDoubleVerts(RHSForm(formula))
 
-    ## DRY ...
+    ## include mustart/etastart in the model frame call (glmer-specific)
```
bbolker: if the only difference here is that we need `mustart` and `etastart` for GLMMs but not LMMs, can we achieve this with an `if (isGLMM)` (or equivalent test) clause?

bbolker (commented):
see comments below (I think we can get a little bit more included in `mkFormula`)

annahuynhly:
I think it's good to review again -- basically, when I ran R CMD check (formally), there were a couple of bugs that weren't catched unrelated to this PR. 

- `glmer.Rnw` used `gridExtra` and `RColorBrewer`, hence added to `DESCRIPTION`
-  needed to get rid of the hard coded `parent.frame(2L)` in `utilities.R` (not compatible with `mkFormula`)
-  changes in `test/testthat/test-methods.R` seem to be because there can be a difference between the number of times the deviance function got called to arrive at the results (which was `function.eval`), which is why it's now ignored in the test

bbolker:
> I think it's good to review again -- basically, when I ran R CMD check (formally), there were a couple of bugs that weren't catched unrelated to this PR.
> 
>   * `glmer.Rnw` used `gridExtra` and `RColorBrewer`, hence added to `DESCRIPTION`

Try merging `master` again (may have to resolve conflicts); `RColorBrewer` isn't really needed explicitly (it's imported by `ggplot2`; already added `gridExtra` (and `see`, `DHARMa`, `performance`)
> 
>  * needed to get rid of the hard coded `parent.frame(2L)` in `utilities.R` (not compatible with `mkFormula`)

OK

> 
>  * changes in `test/testthat/test-methods.R` seem to be because there can be a difference between the number of times the deviance function got called to arrive at the results (which was `function.eval`), which is why it's now ignored in the test

OK

Will look again

Issue #377, opened 2016-05-10
summary.lmList issues


bbolker:
`summary.lmList` can fail when some groups have singular fits.  This is really an upstream problem in `nlme`, since we copy the summary methods for our own use ...

```
p2 <- subset(Pixel, Dog %in% c(7,8,9))
p3 <- subset(Pixel, Dog %in% c(7,8))  ## without problematic case
library("nlme")
L1 <- lmList(pixel~day+I(day^2)|Dog, p2)
try(summary(L1))
L2 <- lme4::lmList(pixel~day+I(day^2)|Dog, p2)
try(summary(L2))
```

bbolker:
It's a built-in data set from the `nlme` package. `?nlme::Pixel`  (and/or `data("Pixel", package = "nlme")`)

>  The ‘Pixel’ data frame has 102 rows and 4 columns of data on the pixel intensities of CT scans of dogs over time

bbolker:
Can we close this now?

annahuynhly:
No; what I solved didn't deal with the issue but some naming issue that I noticed alongside trying to solve this issue. There is still a problem here, that copilot didn't fix (maybe Claude can? I'll put it on my list...)

annahuynhly:
Close; it was solved by GH #970, and for future reference: 'singular fits' is not an appropriate name; should've said "for rank deficient models..."

PR #958, opened 2026-04-02
Fix lme4::lmList prints grouping variable as NULL

annahuynhly:
Every time you use lme4::lmList, for some reason, the print shows that the grouping variable is always NULL. See quick example below:
```
library(lme4)
test <- lme4::lmList(Reaction ~ Days | Subject, sleepstudy)
try(summary(test))
##Call:
##  Model: Reaction ~ Days | NULL 
##   Data: sleepstudy 
```
Note that this does not happen with nlme, which is where the function was copied from:
```
library(nlme)
data(sleepstudy, package="lme4")
test2 <- nlme::lmList(Reaction ~ Days | Subject, sleepstudy)
try(summary(test2))
##Call:
## Model: Reaction ~ Days | Subject 
##  Data: sleepstudy 
```

I found this problem when I was trying to investigate https://github.com/lme4/lme4/issues/377.

---

Now this issue is fixed. The problem was `print.summary.lmList`, which was borrowed from `nlme`, uses `attr(object, "groupsForm")` to pick the name for the `group` part of the `Model:` aspect in `print.summary`. This attribute did get generated before for `lme4::lmList`.

At first I tried to use copilot, but as I quickly had an idea of the issue when I debugged `print.summary.lmList`...

### DIFF
--- a/R/lmList.R
+++ b/R/lmList.R
@@ -176,7 +176,8 @@ lmList <- function(formula, data, family, subset, weights,
         groups = ordered(groups),
         origOrder = match(unique(as.character(groups)), nms)
         )
-    attr(val, "groupsForm") <- reformulate(deparse(mform$groups))
+    # TODO: import reformulas::get_grpvars once reformulas has updated
+    attr(val, "groupsForm") <- reformulate(get_grpvars(formula))
     val
 }
 
--- a/R/lmList.R
+++ b/R/lmList.R
@@ -111,6 +111,11 @@ lmList <- function(formula, data, family, subset, weights,
     ## NA values in unused columns.  All we need the model frame for
     ## is evaluating the groups.
 
+    ff <- reformulas::findbars_x(formula)
+    if (length(ff) > 1) 
+      stop("lme4::lmList only works with simple grouping formulas (not involving * or /); 
+           use interaction(...) to construct an appropriate grouping variable first.")
+    
     ## keep weights and offsets in case we have NAs there??
     m <- match(c("formula", "data", "subset", "na.action",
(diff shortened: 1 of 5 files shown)


bbolker:
`nlme::lmList` evaluates the grouping factors at the *lowest* level, i.e. for `sch/id` it would use the interaction of `sch` and `id` (`sch:id` in formula terms). It's easy enough to use something like `reformulas::expandGrpVar(mform$groups)` to expand `*` and `/` terms, but then we'd also have to figure out how to recover those interactions (in other machinery those formulas are used to construct model matrices via `model.frame()`, but that's not what we need here). `nlme` uses `nlme:::getGroups.data.frame` to construct an interaction, but the code is pretty ugly (pastes together elements from all of the different variables; really should be using `interaction` ...

It would be good to have the printing correct, but I don't think we need to have all of the capabilities of `nlme`'s version. The documentation can tell users to use `interaction()` (possibly with `drop=TRUE`) if they want to do multi-level splitting.

The main point of having a different version of `lmList` in `lme4` is (1) it handles GLMs as well as LMs and (2) it's a little bit more robust than `nlme`'s version, or at least it used to be (they may have fixed some of the issues)

annahuynhly:
Okay, so conclusion:
1. Should correct the printing for sure, using `reformulas` machinery (just a refinement on this pull request)
2. don’t really need to ensure that it works for crossed factors. for nested, propose in the documentation a “work-around” solution using `interaction()`.

and 3. can you respond to my email regarding meeting tomorrow? Happy to meet, but if you celebrate easter or have something planned for the long weekend we can move it.

bbolker:
Add NEWS please ...

Issue #936, opened 2026-03-03
matching up gaussian(link = "log") results across packages


bbolker:
```r
data("Rail", package = "nlme")
library(lme4)
library(glmmTMB)
library(mgcv)
form <- travel~1 + (1|Rail)
m1_glmer <- glmer(form, data=Rail, family=gaussian(link="log"))
m1_glmmTMB <- glmmTMB(form, data=Rail, family=gaussian(link="log"))
m1_mgcv <- gam(travel ~ 1 + s(Rail, bs= "re"), method = "ML", data = Rail, family = gaussian(link="log"))
m1_PQL <- MASS::glmmPQL(travel ~1, random = ~1|Rail, family = gaussian(link = "log"), data = Rail)
```

glmer gets good estimates (apparently) but reports a different sigma (3.33 vs 4.0) and a different log-likelihood (-73 vs -64.5) ... `mgcv` and `glmmPQL` agree pretty well with `glmmTMB` ...  (it is a tiny data set, so ... ?? could be a denominator issue?)

annahuynhly:
Leaving this comment here for me to refer to later. Below is an example where it fails for the `Gamma` family and a log link function (similar to the original example):
```
set.seed(42)
n_groups <- 6      
n_per    <- 3      
b0       <- log(66)  
sigma_re <- log(25)  

group <- rep(1:n_groups, each = n_per)
u     <- rnorm(n_groups, 0, sigma_re)
mu    <- exp(b0 + u[group])
y     <- rgamma(n_groups * n_per,
                shape = 10,         
                rate  = 10/mu)

dat <- data.frame(y = y, group = factor(group))

m_glmer   <- glmer(y ~ 1 + (1|group), data = dat, family = Gamma(link = "log"))
m_glmmTMB <- glmmTMB(y ~ 1 + (1|group), data = dat, family = Gamma(link = "log"))
m_mgcv    <- gam(y ~ 1 + s(group, bs = "re"), data = dat,
                 family = Gamma(link = "log"), method = "ML")
m_PQL     <- MASS::glmmPQL(y ~ 1, random = ~1|group,
                           family = Gamma(link = "log"), data = dat)
```
produces:
```
> sigma(m_glmer)
[1] 0.4898151
> sigma(m_glmmTMB)
[1] 0.3446698
> sigma(m_mgcv)
[1] 0.324375
> sigma(m_PQL)
[1] 0.323956
```

bbolker:
Possibly related to #643 !

[...]

PR #965, opened 2026-04-26
Fixing gaussian(link = "log") issue across packages

annahuynhly:
See: https://github.com/lme4/lme4/issues/936 

Context: I tested out copilot before (disaster) but I wanted to try out Claude Code and the solution seems to actually work well for this issue. The main doubt I have is exactly where they came up with the equation to optimize and whether using built in `optimize` is generally good to do here.

## Background: why sigma was wrong for Gaussian GLMMs                            
   
For a Gaussian GLMM with non-identity link (e.g. `gaussian(link="log")`), `lme4` fits the model via a Laplace approximation, with the Laplace deviance being:

https://github.com/lme4/lme4/blob/3eb680591593c7307c71e969700b0fceca3be6c9/src/respModule.cpp#L161-L163

The main issue is that `ldL2` also depends on `sigmaML`, but this is not used when profiling. 
https://github.com/lme4/lme4/blob/3eb680591593c7307c71e969700b0fceca3be6c9/src/predModule.cpp#L159
(Allegedly, `ldL2` = log det(Λ'Z' diag(μ²/σ²) ZΛ + I)... I saw that `chm_factor_ldetL2` is from the `Matrix` package but I couldn't exactly understand the code.)

More detailed explanation (after prompting)

### Step 1: The GLMM Cholesky factor depends on the working weights W                                 

At PIRLS convergence, the Cholesky factor satisfies (see equation 45 https://lme4.github.io/lme4/articles/Theory.pdf):
image
                
`ldL2` = log|L_Z|² inherits whatever is in `W`.                              
                                                                                                                                    
### Step 2: The profiled likelihood has a |W| term                            
                                                                                
Profiled log-likelihood for a weighted GLMM (equation 46 of same paper):
image
                                                      
### Step 3: For a Gaussian GLMM, W depends on σ
                                             
W is computed as a function of σ. See:

https://github.com/lme4/lme4/blob/3eb680591593c7307c71e969700b0fceca3be6c9/src/respModule.cpp#L179-L183
                    
Therefore we must optimize based on the un-profiled form. `ldL2` was computed at whatever W was at PIRLS convergence which is not the same estimation of sigma that `aic()` used. 
                                                         
### The fix

After the outer optimiser converges at (θ*, β*), a one-dimensional optimisation (using `optimize()`) over σ finds the true Laplace profile MLE by minimising this version of the deviance:                                                                     

D(σ) = ldL2(σ) + ‖u‖² + n·log(2πσ²) + RSS/σ²                           
                  
What I personally find confusing is... where exactly did they come up with this version of the approximation of the deviance?


annahuynhly:
*Check: is this also wrong for gamma where link="log" as well?  (The log link shouldn't be the relevant issue - I'm expecting the problem to occur for any GLMM with a non-fixed scale parameter (e.g. gaussian, Gamma, ...) - i.e. should happen with link = "inverse" too, although I don't see anything wrong with doing the primary testing on the link="log" case

annahuynhly:
* just realized that I also didn't even add any tests!!!

annahuynhly:
Adding comments from our meeting:
1. it's a weird choice to make changes in `mkMerMod`; ideally, if we want to change the calculations it should be elsewhere
2. it opting to compute using the likelihood instead of `pwrss` is a bit confusing given that `pwrss` is supposedly more robust (but it could be the case that this is the reason it does not match the others...)

bbolker:
think this is resolved (mostly??) by the Gamma_GLMM related work (fixing dispersion estimates for GLMMs with free parameters). Need to check back (see `misc/Gamma_GLMM/README.md`)

Issue #962, opened 2026-04-16
glmer + ar1 returns zero values for fixed effect coefficients


bbolker:
via Jim Hughes on r-sig-mixed. Log-likelihoods, VarCorr, etc. all match reasonably well between `glmmTMB` and `lme4`, so the model itself must be working OK - something must be going wrong downstream with parameter assignment ... ??  In contrast, there's no problem with `lmer`.  Something is wrong with `lme4_1@pp$beta(1)`, which is used to get fixed effect coefficients ... ?? 

```r
form <- out_bin_1 ~ ftime + trt + ar1(ftime + 0 | id_cluster) +
  (1 | id_individual)

library(glmmTMB)
library(lme4)
set.seed(101)
sfun <- function(n) factor(sample(1:n, replace=TRUE, size = 1465))
dd <- data.frame(id_cluster = sfun(7), id_individual = sfun(486),
                 ftime = sfun(4), trt = sfun(2))
np <- list(beta = c(0, rep(0.1, 3), 1), theta = rep(0, 3))
dd$out_bin_1 <- simulate_new(form[-2], newdata = dd,
             newparams = np,
             family = binomial)[[1]]

lme4_1 <- glmer(form, data = dd, family = binomial)
glmmTMB_1 <- glmmTMB(form, data = dd, family = binomial)
c(logLik(lme4_1) - logLik(glmmTMB_1))
logLik(lme4_1)
logLik(glmmTMB_1)

fixef(lme4_1)
lme4_1@optinfo$val

dd$out_gauss <- simulate_new(form[-2], newdata = dd,
             newparams = c(np, list(betadisp=0)),
             family = gaussian)[[1]]

form2 <- update(form, out_gauss ~ .)
lme4_2 <- lmer(form2, data = dd, REML = FALSE)
glmmTMB_2 <- glmmTMB(form2, data = dd, family = gaussian)
fixef(lme4_2)
```

PR #966, opened 2026-05-07
Fixing the bug in GH962

annahuynhly:
See: https://github.com/lme4/lme4/issues/962

I fear Claude code is getting too good.

I did not write NEWS or excessive tests yet; I just wanted @bbolker to see what the solution was for an opinion first, out of request of an experiment...

If approved: will add tests in, including the ./misc file of other experiments for stretch testing.

```
> form <- out_bin_1 ~ ftime + trt + ar1(ftime + 0 | id_cluster) +
+   (1 | id_individual)
> library(lme4)
ℹ Loading lme4
Loading required package: Matrix
> library(glmmTMB)
> set.seed(101)
> sfun <- function(n) factor(sample(1:n, replace=TRUE, size = 1465))
> dd <- data.frame(id_cluster = sfun(7), id_individual = sfun(486),
+                  ftime = sfun(4), trt = sfun(2))
> np <- list(beta = c(0, rep(0.1, 3), 1), theta = rep(0, 3))
> dd$out_bin_1 <- simulate_new(form[-2], newdata = dd,
+                              newparams = np,
+                              family = binomial)[[1]]
> lme4_1 <- glmer(form, data = dd, family = binomial)
> glmmTMB_1 <- glmmTMB(form, data = dd, family = binomial)
> c(logLik(lme4_1) - logLik(glmmTMB_1))
[1] -1.655188e-05
> logLik(lme4_1)
'log Lik.' -818.4017 (df=8)
> logLik(glmmTMB_1)
'log Lik.' -818.4017 (df=8)
> VarCorr(lme4_1)
 Groups        Name        Std.Dev. Corr        
 id_individual (Intercept) 1.1007               
 id_cluster    ftime1      1.1752   -0.153 (ar1)
> fixef(lme4_1)
(Intercept)      ftime2      ftime3      ftime4        trt2 
  0.2807946   0.3508938   0.3081112  -0.8380135   1.1449089 
> lme4_1@optinfo$val
[1]  1.1006869  1.1751912 -0.1525769  0.2807946  0.3508938  0.3081112 -0.8380135
[8]  1.1449089

> dd$out_gauss <- simulate_new(form[-2], newdata = dd,
+                              newparams = c(np, list(betadisp=0)),
+                              family = gaussian)[[1]]
> form2 <- update(form, out_gauss ~ .)
> lme4_2 <- lmer(form2, data = dd, REML = FALSE)
> glmmTMB_2 <- glmmTMB(form2, data = dd, family = gaussian)
> fixef(lme4_2)
(Intercept)      ftime2      ftime3      ftime4        trt2 
-0.13939738 -0.44396444 -0.01480478  0.27960518  1.03163687 
```


R/utilities.R
```diff
@@ -515,7 +515,11 @@ mkMerMod <- function(rho, opt, reTrms, fr, mc, lme4conv=NULL) {
     sigmaML <- pwrss/n
     if (rcl != "lmerResp") {
         pars <- opt$par
-        if (length(pars) > length(pp$theta)) beta <- pars[-(seq_along(pp$theta))]
+        ## For structured covariance models (e.g., ar1), the optimizer works in a
+        ## lower-dimensional par-space (length(reTrms$lower)) while pp$theta lives
+        ## in the larger theta-space, so use reTrms$lower to count theta params.
+        npar <- length(reTrms$lower)
+        if (length(pars) > npar) beta <- pars[-seq_len(npar)]
```
bbolker: This would be better if we could use `if (length(pars) > getParLength(object))` (or, `npar <- getParLength(object)` since we have to use `npar` twice in the next line; is there a way to avoid referring to `npar` twice without being too clever? more fundamentally, I wonder if `length(npars) > npar` iff `nAGQ > 0`, in which case that would be the test we should use ...), but we don't have access to a `merMod` object at this point (this is within `mkMerMod`, where we're trying to *build* the `merMod` object; the `getParLength()` method for `merMod` objects uses `length(object@lower)` anyway, and at the end of this function, we assign `reTrms$lower` to the `@lower` slot of the returned value). 

It would be slightly better, I think, to include an element `npar = length(reTrms$lower)` in the `dims` list above (i.e. line 502), and possibly to include it as an option in `getME()` (there are some very mild backward compatibility issues here, as we would be changing the structure of the stored object, but I think it would be worth it). (The elements of `dims` are described in `man/getME.Rd`, this should be updated as well.) Once this change was made we would use `dims[["npar"]]` here (and anywhere else we need to know this piece of information).  If you understand what I'm saying here you can make the changes, otherwise I will ...

bbolker (changes requested):

bbolker:
I'm actually perfectly happy that Claude found this. It's not conceptually deep ("obvious in hindsight" that this is the problem), and it saved a lot of debugging time. Hopefully it will fix all of the issues we've found so far (some of the examples of structured-covariance models for the Contraception data at the end of the `glmer` paper, that were previously not run, are also acting a little wonky, hopefully (?) it's a manifestation of the same issue ...

R/utilities.R
annahuynhly: I'll make the change myself; I think I understand what you're saying and I'll do it alongside incorporating multiple tests.

annahuynhly:
Okay, will be doing the following next (finish by tonight):
1. add tests
2. update NEWs
3. make the slight adjustment that you recommended

Will make another comment when complete and needs to be reviewed.

bbolker:
Great. I edited my review a little bit. I don't know if it's worth trying to be clever and figure out/be sure that `length(pars) > npar)` is exactly equivalent to `nAGQ > 0`  ... ? I can't *think* of any other case ... maybe we could code it as `nAGQ > 0` and put in an assertion, e.g.

```
if ((length(pars) == dims[["npar"]]) != (nAGQ==0)) stop("something surprising happened")
```
this wouldn't cost much computation time and could save us headaches if there's a case we forgot about (obviously the error message should be more informative than that)

bbolker:
PPS if some of the tests are slow you might want to include a `testLevel` clause (e.g. `skip_if(testLevel < 2)`)

annahuynhly:
Ugh, this took longer than expected (still not done) because now there's a big issue with `nlmer` modes.

In particular, this new implementation won't work for `nlmerResp` as well BECAUSE `dims[["nAGQ"]]` is `null`. However, cureently the code only says if `rcl != "lmerResp"`... Can I just write `rcl = "glmerResp"`? Or should it be the case that if we have `nlmerResp` then we have the original code before hand (as an `if else` statement...)
```
if (rcl != "lmerResp") {
        pars <- opt$par
        ## making the assertion that length(npars) > npar iff nAGQ > 0,
        ## error code below is for the case that it fails
        if ((length(pars) == dims[["npar"]]) != (dims[["nAGQ"]]==0L))
            stop(sprintf("unexpected parameter vector length: length(pars)=%d, npar=%d, nAGQ=%d; ",
                         length(pars), dims[["npar"]], dims[["nAGQ"]]),
                 "expected length(pars)==npar iff nAGQ==0")
        ## For structured covariance models (e.g., ar1), npar = length(reTrms$lower)
        ## may be less than length(pp$theta), so we use dims[["npar"]] to index correctly.
        if (dims[["nAGQ"]] > 0L) beta <- pars[-seq_len(dims[["npar"]])]
    }
```
I'm just clarifying as I feel like my proposed solution sounds a bit of a "monkey" coded solution, but I don't still understand `nlmer` and I know it has fell through the cracks... multiple times...

bbolker:
**tl;dr** I think we have to go back and make sure `nAGQ` is set properly for `nlmer` models.

I think this is an oversight in the setup of `nlmer`. `dims[["nAGQ"]]` is set from `rho$nAGQ`.  The `rho` environment is populated [here](https://github.com/lme4/lme4/blob/master/R/lmer.R#L227-L246) for `nlmer`, and `nAGQ` isn't included. In contrast, `rho$nAGQ` gets set [here](https://github.com/lme4/lme4/blob/master/R/modular.R#L989) and [here](https://github.com/lme4/lme4/blob/master/R/modular.R#L898) for `glmer` models ...

R/lmer.R
```diff
@@ -2131,6 +2132,7 @@ getME.merMod <- function(object,
            "n" = dims[["n"]],
            "p" = dims[["p"]],
            "q" = dims[["q"]],
+           "npar" = if ("npar" %in% names(dims)) dims[["npar"]] else length(getLower(object)),
```
annahuynhly: *Leaving a comment here since it's weird: basically, for old `rda` files we have that `dims[["npar"]]` does not exist. This solution here is to help with backwards compatibility.

annahuynhly:
@bbolker ready to be reviewed again.

bbolker:
I think that

```r
if ("npar" %in% names(dims)) dims[["npar"]] else length(getLower(object))
```

can be written more compactly as 

```r
as.list(dims)[["npar"]] %||%  length(getLower(object))
```

* `as.list()`: converts `dims` to a list so that `...[[<non-existent>]]` will return NULL rather than an out-of-bounds error
* the `%||%` operator (introduced in R 4.4.0, but defined backward compatibly in `R/zzz.R`) is short hand for "substitute if NULL", i.e. `a %||% b` is equivalent to `if (is.null(a)) b else a`

annahuynhly:
Anything else?

bbolker:
Not especially, but I'll have to look it over tomorrow ...

bbolker:
didn't actually intend to merge this (I think it happened indirectly by merging into `glmer_JSS_add` and then merging that into master??) but I'm satisfied that I did ...

Issue #968, opened 2026-05-13
Douglas Bates' Comments on the glmer JSS (?) paper


annahuynhly:
Some of the comments from the email thread; pasting here on behalf of @dmbates 

~~> I'm not sure that I should be listed as first author as I really haven't contributed to the paper or the lme4 package for many years.  Also, as a retiree padding my resumé is not as important as it was in the past.~~
New order (for now; can be changed): Anna, Steven, Rune, Doug, Martin, then Ben

> I too am surprised with the results of the profiling being so different from the parametric bootstrap and Wald intervals in a couple of cases.  For the fixed-effects parameters the profiling can be straightforwardly implemented by incorporating the contribution of the parameter being profiled into the offset.   At present we don't have profiling implemented for GLMMs in MixedModels.jl but I could see if I can cobble something together to check those results.
EDIT: please see results here: https://github.com/lme4/lme4/pull/969#issuecomment-4465202951

> Regarding the MixedModels.jl approach, we have decided to submit the "Blocked Cholesky" manuscript to J. Data Science instead of J. Statist. Soft. because it seems like it will be a massive hassle to get around their "results must be reproducible to the last significant digit" policy.  So it may be in a citable form if I can motivate myself to finish the modifications to the manuscript.
NOTE: not too important to deal with right now; mostly just an issue if we submit to JSS...

~~> I'm not sure if I have mentioned it before but we recently concluded that there is no need to use constrained optimization on the profiled log-likelihood for LMMs and GLMMs.  You can establish that the log-likelihood is an even function of the parameters that occur on the diagonal of Λ, which means that the optimizer should converge to values very close to zero in the case of a singular r.e. covariance matrix.  The reason I mention this is not because I am dying to twerk the lme4 code but because it may make it possible to choose from a broader selection of optimizers.~~

> ~~eq 22 alignment problem~~
~~Minor point but the GH weights and locations on p. 9 could benefit by averaging the weights in forward and reverse order and the locations in forward order and negative locations in reverse order.~~
~~p. 10 agree that discussion of negative binomial glmm should go elsewhere~~
~~p. 10, sec. 4 We recently realized that there is no need to constrain the elements of θ that are on the diagonal of Λ.  It may be worthwhile mentioning this.~~
~~p. 10 sec. 4 the parenthetical expression starting with "as specified" is not closed.  I think it would be better to write it as "the current default, as specified in glmerControl, is ..."~~
[...]

PR #969, opened 2026-05-13
Changes based on Doug Bates' Comments

annahuynhly:
See: https://github.com/lme4/lme4/issues/968

EDIT: this is ready to be reviewed.


annahuynhly:
# glmmTMB

## cbpp: 
image
 
## contraception:
image
 
---
# using another optimizer ("nloptwrap")

## cbpp:
image

## contraception:
image

annahuynhly:
@bbolker ready to be reviewed!

Also, quick comment: Steve said he doesn't feel comfortable being an author anymore; should we just remove him? Or should we forcibly add him in since he HAS contributed in the past?

bbolker:
I can imagine trying to convince Steve to stay on, but the basic principle is that people shouldn't be authors if they don't want to be.

bbolker:
I've been fussing with this a lot. I think the only thing left to do is figure out if there's a way to make any of the profile CIs more stable (or make excuses for them). From @annahuynhly's example above, `glmmTMB` is doing a better job with these ...

bbolker:
PS **strongly** tempted to add a comment at this point about "profiling is sometimes numerically unstable" and let it go at that ...

Issue #930, opened 2026-03-02
Unnecessary coercion and memory allocations from using logical `NA` as the default


HenrikBengtsson:
For example,

https://github.com/lme4/lme4/blob/a3ef89e32ea207a318e9090c83942ff5878238f8/R/lmList.R#L210-L216

sets up a _logical_ matrix, because it uses `NA` as the default value, which is then coerced to a numeric matrix in the for loop.

Updating to:

```r

 co <- matrix(NA_real_, 
              ncol = length(allnames), 
              nrow = length(coefs), 
              dimnames = list(names(object), allnames)) 
```

avoids the coercion and thereby also save an unnecessary memory allocation and garbage collection.

# Details

Below are other _candidates_ of `NA` that might benefit from using another type, e.g. `NA_real_`;

```sh
$ grep -F "NA," R/*.R
R/bootMer.R:                    use.u = FALSE, re.form = NA,
R/bootMer.R:                                          control = control)), errval = rep(NA, length.t0))
R/deriv.R:                    lower=rep(NA,length(x)), upper=rep(NA,length(x)), ...) {
R/lmer.R:    modp <- (as.logical(vapply(dots, is, NA, "merMod")) |
R/lmer.R:             as.logical(vapply(dots, is, NA, "lm")))
R/lmer.R:        if(!all(vapply(data, identical, NA, data[[1]])))
R/lmer.R:        if(!all(vapply(subset, identical, NA, subset[[1]])))
R/lmer.R:        chisq <- 2 * pmax(0, c(NA, diff(llk)))
R/lmer.R:        dfChisq <- c(NA, diff(npar))
R/lmer.R:                          "Pr(>Chisq)" = ifelse(dfChisq==0,NA,pchisq(chisq, dfChisq, lower.tail = FALSE)),
R/lmer.R:        xx <- rep(NA,nrow(model.frame(object)))
R/lmer.R:             pwrss=pwrss, drsum=NA, dev=opt$fval, REML=NA,
R/lmer.R:               getNBdisp(object) else NA,
R/lmer.R:                      error=function(e) matrix(NA,nrow=nrow(h),ncol=ncol(h)))
R/lmer.R:        rr <- matrix(NA,nrow(V),ncol(V))
R/lmList.R:  errs <- vapply(x, inherits, NA, what = "error")
R/lmList.R:            co <- matrix(NA,
R/predict.R:            model.frame(tt, newdata.NA, na.action=na.pass, xlev=orig.random.levs))
[...]

PR #974, opened 2026-05-16
NA_real_ takes over our life

annahuynhly:
This is to fix the issue: https://github.com/lme4/lme4/issues/930 

I also asked Claude to point out more places where we could replace `NA` with `NA_real`. Some of the suggestions did break tests, so I left some `NA` alone.

Verified with `R CMD check`.

### DIFF
--- a/R/bootMer.R
+++ b/R/bootMer.R
@@ -29,7 +29,7 @@
 ##' The semi-parametric variant is not yet implemented, and we only
 ##' provide a method for \code{\link{lmer}}  and \code{\link{glmer}} results.
 bootMer <- function(x, FUN, nsim = 1, seed = NULL,
-                    use.u = FALSE, re.form = NA_real_,
+                    use.u = FALSE, re.form = NA,
                     type = c("parametric","semiparametric"),
                     verbose = FALSE,
                     .progress = "none", PBargs=list(),
@@ -108,7 +108,7 @@ bootMer <- function(x, FUN, nsim = 1, seed = NULL,
       control
       length.t0 <- length(t0)
       f1 <- factory(function(i) FUN(refit(x, ss[[i]],
-                                          control = control)), errval = rep(NA_real_, length.t0))
+                                          control = control)), errval = rep(NA_character_, length.t0))
       function(i) {
           ret <- f1(i)
           if (verbose) { cat(sprintf("%5d :",i)); str(ret) }
--- a/R/bootMer.R
+++ b/R/bootMer.R
@@ -29,7 +29,7 @@
 ##' The semi-parametric variant is not yet implemented, and we only
 ##' provide a method for \code{\link{lmer}}  and \code{\link{glmer}} results.
 bootMer <- function(x, FUN, nsim = 1, seed = NULL,
-                    use.u = FALSE, re.form = NA,
+                    use.u = FALSE, re.form = NA_real_,
                     type = c("parametric","semiparametric"),
                     verbose = FALSE,
                     .progress = "none", PBargs=list(),
@@ -108,7 +108,7 @@ bootMer <- function(x, FUN, nsim = 1, seed = NULL,
       control
       length.t0 <- length(t0)
       f1 <- factory(function(i) FUN(refit(x, ss[[i]],
-                                          control = control)), errval = rep(NA, length.t0))
+                                          control = control)), errval = rep(NA_real_, length.t0))
       function(i) {
           ret <- f1(i)
           if (verbose) { cat(sprintf("%5d :",i)); str(ret) }
--- a/R/lmList.R
+++ b/R/lmList.R
@@ -47,7 +47,7 @@ if(getRversion() < "3.5.0") {
 ##' Utility for lmList(), ...: Collect errors from a list \code{x},
 ##' produce a "summary warning" and keep that message as "warningMsg" attribute
 warnErrList <- function(x, warn = TRUE) {
-  errs <- vapply(x, inherits, NA_real_, what = "error")
+  errs <- vapply(x, inherits, NA, what = "error")
   if (any(errs)) {
     v.err <- x[errs]
     e.call <- paste(deparse(conditionCall(v.err[[1]])), collapse = "\n")
--- a/R/lmList.R
+++ b/R/lmList.R
@@ -47,7 +47,7 @@ if(getRversion() < "3.5.0") {
 ##' Utility for lmList(), ...: Collect errors from a list \code{x},
 ##' produce a "summary warning" and keep that message as "warningMsg" attribute
 warnErrList <- function(x, warn = TRUE) {
-  errs <- vapply(x, inherits, NA, what = "error")
+  errs <- vapply(x, inherits, NA_real_, what = "error")
   if (any(errs)) {
     v.err <- x[errs]
     e.call <- paste(deparse(conditionCall(v.err[[1]])), collapse = "\n")
@@ -216,7 +216,7 @@ coef.lmList4 <- function(object,
         ## different parameter sets may be estimated for different subsets of data ...
         allnames <- Reduce(union, lapply(coefs[non.null], names))
         if (is.numeric(template)) {
-            co <- matrix(NA,
+            co <- matrix(NA_real_,
                          ncol = length(allnames),
                          nrow = length(coefs),
                          dimnames = list(names(object), allnames))
--- a/R/lmer.R
+++ b/R/lmer.R
@@ -511,8 +511,8 @@ anovaLmer <- function(object, ..., refit = TRUE, model.names=NULL) {
     mCall <- match.call(expand.dots = TRUE)
     dots <- list(...)
     .sapply <- function(L, FUN, ...) unlist(lapply(L, FUN, ...))
-    modp <- (as.logical(vapply(dots, is, NA, "merMod")) |
-             as.logical(vapply(dots, is, NA, "lm")))
+    modp <- (as.logical(vapply(dots, is, NA_real_, "merMod")) |
+             as.logical(vapply(dots, is, NA_real_, "lm")))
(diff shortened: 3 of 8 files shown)


man/lmer.Rd
```diff
@@ -92,7 +92,7 @@ lmer(formula, data = NULL, REML = TRUE, control = lmerControl(),
     deviation parameter(s) followed by the correlation parameter
     \eqn{\rho} (specified on the [0,1] scale).
   }
-  See the \href{../doc/covariance_structures.html}{covariance structures vignette} for more detail on covariance matrix
```
bbolker, line 95: why does this keep getting removed? I'd like to have the link there. Does it not work on your system ... ?? Or does Claude not like it .. .??

R/bootMer.R
```diff
@@ -29,7 +29,7 @@
 ##' The semi-parametric variant is not yet implemented, and we only
 ##' provide a method for \code{\link{lmer}}  and \code{\link{glmer}} results.
 bootMer <- function(x, FUN, nsim = 1, seed = NULL,
-                    use.u = FALSE, re.form = NA,
+                    use.u = FALSE, re.form = NA_real_,
```
bbolker: I don't think this should be `NA_real_`. `re.form` is a flexible argument, which can be either NULL, NA, or a formula. It makes most sense for the `NA` placeholder value to be logical (which is what naked/default `NA` is: `typeof(NA)` == "logical"

R/bootMer.R
```diff
@@ -108,7 +108,7 @@ bootMer <- function(x, FUN, nsim = 1, seed = NULL,
       control
       length.t0 <- length(t0)
       f1 <- factory(function(i) FUN(refit(x, ss[[i]],
-                                          control = control)), errval = rep(NA, length.t0))
+                                          control = control)), errval = rep(NA_real_, length.t0))
```
bbolker: If not plain `NA`, this should be `NA_character`.

R/lmer.R
```diff
@@ -511,8 +511,8 @@ anovaLmer <- function(object, ..., refit = TRUE, model.names=NULL) {
     mCall <- match.call(expand.dots = TRUE)
     dots <- list(...)
     .sapply <- function(L, FUN, ...) unlist(lapply(L, FUN, ...))
-    modp <- (as.logical(vapply(dots, is, NA, "merMod")) |
-             as.logical(vapply(dots, is, NA, "lm")))
+    modp <- (as.logical(vapply(dots, is, NA_real_, "merMod")) |
```
bbolker: this should be `NA`. This might make more sense if the arguments were named, i.e. `vapply(dots, FUN = is, FUN.VALUE = NA, "merMod")` . Not sure why we need the `as.logical` in any case as (especially if we use `NA`) `modp` should be a logical vector in the first place ... ?? Look at `git blame` to see who wrote this in the first place ... ?

R/lmer.R
```diff
@@ -936,7 +936,7 @@ fitted.merMod <- function(object, ...) {
     xx <- object@resp$mu
     if (length(xx)==0) {
         ## handle 'fake' objects created by simulate()
-        xx <- rep(NA,nrow(model.frame(object)))
+        xx <- rep(NA_real_,nrow(model.frame(object)))
     }
     if (is.null(nm <- rownames(model.frame(object)))) nm <- seq_along(xx)
```
bbolker: while we're at it this could be replaced by 

```
nm <- rownames(model.frame(object)) %||% seq_along(xx)
```

R/lmer.R
```diff
@@ -1593,7 +1593,7 @@ refitML.merMod <- function (x, optimizer="bobyqa", ...) {
     ussq <- pp$sqrL(1)
     pwrss <- wrss + ussq
     cmp <- c(ldL2=pp$ldL2(), ldRX2=pp$ldRX2(), wrss=wrss, ussq=ussq,
-             pwrss=pwrss, drsum=NA, dev=opt$fval, REML=NA,
+             pwrss=pwrss, drsum=NA_real_, dev=opt$fval, REML=NA_real_,
```
bbolker: `REML` value should definitely be `NA`, not `NA_real`. Not sure about `drsum` without checking.

R/lmList.R
```diff
@@ -47,7 +47,7 @@ if(getRversion() < "3.5.0") {
 ##' Utility for lmList(), ...: Collect errors from a list \code{x},
 ##' produce a "summary warning" and keep that message as "warningMsg" attribute
 warnErrList <- function(x, warn = TRUE) {
-  errs <- vapply(x, inherits, NA, what = "error")
+  errs <- vapply(x, inherits, NA_real_, what = "error")
```
bbolker: see parallel comments about `vapply` above, I think this should be `FUN.VALUE = NA`

R/predict.R
```diff
@@ -291,7 +291,7 @@ mkNewReTrms <- function(object, newdata,
         ## ReTrms <- reformulas::mkReTrms(reformulas::findbars(re.form[[2]]), rfd)
         ## update Lambdat (ugh, better way to do this?)
         reTrms <- within(reTrms,Lambdat@x <- unname(getME(object,"theta")[Lind]))
-        if (!allow.new.levels && any(vapply(reTrms$flist, anyNA, NA)))
+        if (!allow.new.levels && any(vapply(reTrms$flist, anyNA, NA_real_)))
```
bbolker: should be `NA` I think

R/predict.R
```diff
@@ -671,7 +671,7 @@ simulate.formula_lhs_matrix <- simulate.formula_lhs_numeric <-
         }
 
 simulate.merMod <- function(object, nsim = 1, seed = NULL, use.u = FALSE,
-                            re.form=NA,
+                            re.form=NA_real_,
```
bbolker: should be `NA`

man/lmer.Rd
bbolker: I suspect the link might be broken in a `devtools` context, but not if you do a full build + install ... ?

bbolker:
A large part of this is Claude being over-eager. See my comments throughout (I didn't finish reviewing). I don't know if you can re-prompt Claude to go back and fix its errors or not ... ?? (This pull request is *almost* a waste of time - there is a point to cleaning up little inconsistencies, but Claude got it wrong probably more than half the time ...)

annahuynhly:
I don't think your review went through -- I should be able to easily fix the ones that Bengtsson suggested vs the ones that Claude suggested though. Commit [f448f13](https://github.com/lme4/lme4/pull/974/commits/f448f130bea0a126882b229b76fc4fdf425d401d) is actually specifically from Claude's comments, so I can just revert. (My bad for wasting your time!)

bbolker:
Figured out which button to press to submit my review.  Not all of Bengtsson's suggestions are correct, either. You should be able to get the idea from my reviews of what the problems are/which kinds of `NA`s should be left alone vs converted to `NA_real_`.  As I may have mentioned, I wonder if it's possible to prompt Claude to be a little bit more discriminating ... ? (It's plausible that it's within its capabilities to determine whether an `NA` should be logical or not within its local context ...)

man/lmer.Rd
annahuynhly: Oh, I've been removing it manually since like you said, it was broken in a `devtools` context. I'll restore it

annahuynhly:
I'll handle it -- I know some of them were incorrect as they were flagged by the tests, but I was probably also overly eagar with getting rid of `NA`. Will report back when modified

annahuynhly:
Okay @bbolker, hopefully not a waste of time for you to review now...

R/lmer.R
```diff
@@ -587,7 +587,7 @@ anovaLmer <- function(object, ..., refit = TRUE, model.names=NULL) {
                           "-2*log(L)" = -2*llk,
                           Chisq = chisq,
                           Df = dfChisq,
-                          "Pr(>Chisq)" = ifelse(dfChisq==0,NA,pchisq(chisq, dfChisq, lower.tail = FALSE)),
+                          "Pr(>Chisq)" = ifelse(dfChisq==0,NA_real_,pchisq(chisq, dfChisq, lower.tail = FALSE)),
```
bbolker: I think this will actually be `NA_integer` (since difference of number of parameters is always integer; in principle fractional df are possible, but not in this context ...)

R/predict.R
```diff
@@ -732,7 +732,7 @@ simulate.merMod <- function(object, nsim = 1, seed = NULL, use.u = FALSE,
             devfun <- do.call(mkLmerDevfun, lmod)
             object <- mkMerMod(environment(devfun),
                                ## (real parameters will be filled in later)
-                               opt = list(par=NA,fval=NA,conv=NA),
+                               opt = list(par=NA_real_,fval=NA_real_,conv=NA_real_),
```
bbolker: I think `conv` might be `NA_integer` (convergence code)?

R/predict.R
```diff
@@ -743,7 +743,7 @@ simulate.merMod <- function(object, nsim = 1, seed = NULL, use.u = FALSE,
             devfun <- do.call(mkGlmerDevfun, glmod)
             object <- mkMerMod(environment(devfun),
                                ## (real parameters will be filled in later)
-                               opt = list(par=NA,fval=NA,conv=NA),
+                               opt = list(par=NA_real_,fval=NA_real_,conv=NA_real_),
```
bbolker: ditto, see above (this will interact with the "lFormula/glFormula" PR/changes - we shouldn't have to repeat this ...)

bbolker (changes requested):
Only a couple of fussy bits left to fix (`NA_real` → `NA_integer` in a few places)

annahuynhly:
Okay; I'll also double check if anything else needs to be an integer too later. Will @ you again once done.

annahuynhly:
@bbolker ?

PR #977, opened 2026-05-29
Minor bug in tests occur when devdiff is NULL

annahuynhly:
Sometimes `devdiff` is passed as `NA_real_` in which there is an error with the line `if(devdiff <= 0)`...

### DIFF
--- a/R/profile.R
+++ b/R/profile.R
@@ -263,7 +263,7 @@ profile.merMod <- function(fitted,
                 if(devdiff < 0)
                     warning(gettextf("slightly lower deviances (diff=%g) detected",
                                      devdiff), domain=NA)
-                if (devdiff <= 0) devdiff <- devtol
+                if (is.na(devdiff) && devdiff <= 0) devdiff <- devtol
             }
             zz <- sign(xx - pw) * sqrt(devdiff)
             r <- c(zz, mkpar(npar1, w, xx, pars))


annahuynhly:
@bbolker not trying to be too annoying; but I just want to double check this before I do check reverse....

bbolker:
I feel like I resolved the problem you're talking about in b913b3d52349250f (but, alas, I failed to include a test in that commit ...)  Can you confirm that the problem occurs despite this  change?

(NULL and NA_real_ are two different values, so the title of the issue seems misleading?)

bbolker:
I'm having trouble seeing how an NA can occur here given that the relevant line is inside an `else` clause of an `if (is.na(devdiff))` branch ... ??? 

```r
   if (is.na(devdiff)) {
                warning("NAs detected in profiling")
                ## optimizer failed; return an all-NA row so fillmat can skip it
                return(rep(NA_real_, ncol(res)))
} else {
                if(verbose && devdiff < 0)
                    cat("old deviance ",base,",\n",
                        "new deviance ",ores$fval,",\n",
                        "new params ",
                        paste(mkpar(npar1,w,xx,ores$par),
                              collapse=","),"\n")
                
                if (devdiff < (-devtol))
                    stop("profiling detected new, lower deviance ",
                         sprintf("(deviance diff = %1.3g, tolerance = %1.3g)",
                                 abs(devdiff), devtol))
                             
                if(devdiff < 0)
                    warning(gettextf("slightly lower deviances (diff=%g) detected",
                                     devdiff), domain=NA)
                if (devdiff <= 0) devdiff <- devtol
}
```

annahuynhly:
Closing; what you're saying seems perfectly reasonable. I must've messed up earlier when trying to update my fork somehow...

Issue #913, opened 2026-02-11
unused control arguments warning when switching optimizers for profiling


bbolker:
```r
library(lme4)
fm1 <- lmer(Reaction ~ Days + (Days | Subject), sleepstudy)
p <- profile(fm1, which = 1, optimizer = "bobyqa")
```

20 warnings of the type

> In (function (npt = min(n + 2L, 2L * n), rhobeg = NA,  ... : unused control arguments ignored

This comes ultimately from `minqa:::commonArgs()`, when `print_level` is in the `control` argument.

I guess this gets added to control when we use optimizer "nloptwrap" in the first call, then passed along to profile even if we change optimizers?

annahuynhly:
This got closed but my PR wasn't approved yet..?

bbolker:
fixed now

PR #981, opened 2026-06-19
Fixing the bug in issue913

annahuynhly:
*Might need to look at some of the issues when I did `R CMD check`... but just look at it for now

See: https://github.com/lme4/lme4/issues/913

Most of my issues were due to not being able to build the vignettes, so this is probably fine.


annahuynhly:
@bbolker is this good? The issue got closed

bbolker:
thanks, forgot to merge (!)

PR #982, opened 2026-06-22
Change organization of topics in glmer vignette

dmbates:
The changes here are to the glmer vignette, to describe a GLMM as derived from an LM through adding random effects, as in an LMM, and mapping the linear predictor to the mean response with a possibly non-Gaussian distribution family, as in a GLM.

I noticed something peculiar in eq. 10 (around line 133).  The minimization is with respect to $\bm\delta$ but $\bm\delta$ doesn't explicitly occur in the expression being optimized.  I think maybe it should be $\bm\delta$ instead of $\bm u$ in that expression.  I have made that mistake in the past of evaluating an increment to the next position (using the weighted, working residual instead of the weighted, working response) rather than evaluating the next position, only to discover that the penalty is on the length of $\bm u$, not the length of $\bm\delta$.

Fixing this may be as simple as writing $\bm u$ instead of $\bm\delta$ for the coefficients being optimized.  Alternatively, it may be that $\bm\delta$ should be used instead of $\bm u$ in the expression to be optimized.  I am beginning to think that the second one is the correct approach as the penalty is on $\bm u_{i+1} = \bm u_i + \bm\delta$.

In any case, we might want to brainstorm about a notation to indicate the value at the t'th iteration.  The problem with using a subscript i is that we are using i as an index of the observation, as in $y_i, i=1,\dots,n$.  I should probably roll up my sleeves and read the code to see what is actually done.

I pointed out previously that we don't define $\bm P$ in equations (13) and (14).  It's a question of how explicit we need to be about solving the penalized least squares problem.


bbolker:
I've defined $\boldsymbol P$ (l. 152), and introduced the notation $u^{(i)}$ for (e.g.) the value of $u$ at the $i^\textrm{th}$ PIRLS iteration.  (I think it's OK to use $i$ as the index here rather than, say, $k$, because the other uses of $i$ to index elements of a vector ($y_i$) are fairly distant in the text.  But it might not hurt to use $k$ instead ...?)

Two outstanding issues:

* still not quite sure what to do about making equation 10 correct/consistent (i.e. whether we should be taking argmin of $\delta$ or $u$, or whether $\delta$ or $u$ should appear in the equation. FWIW the actual solver code from `external.cpp` is:

```cpp
    merPredD::Scalar merPredD::solveU() {
        d_delb.setZero(); // in calculation of linPred delb should be zero after solveU
        d_delu          = d_Utr - d_u0;
        d_L.solveInPlace(d_delu, CHOLMOD_P);
        d_L.solveInPlace(d_delu, CHOLMOD_L);    // d_delu now contains cu
        d_CcNumer       = d_delu.squaredNorm(); // numerator of convergence criterion
        d_L.solveInPlace(d_delu, CHOLMOD_Lt);
        d_L.solveInPlace(d_delu, CHOLMOD_Pt);
  return d_CcNumer;
    }
```

My current best guess is that we should write this out as

Screenshot from 2026-06-23 09-35-54

(sorry for image - Github doesn't render complicated LaTeX nicely), *or* explicitly replace $u^{(i+1)}$ with $u^{(i)} + \delta^{(i)}$ ... ?


* the other issue is one of PIRLS-step indexing, i.e. whether we're writing things out as going from step $i-1$ to $i$ or from $i$ to $i+1$. Around Eq. 10 we're framing things as going from $i$ to $i+1$, while farther down (138ff, step-halving discussion) we're going from $i-1$ to $i$ ...

dmbates:
Did you mean to have $u^{(i)}$, not $u^{(i+1)}$ in the left summand within the squared norm expression in that image?  This has $u^{(i+1)}$ in both the left and the right summands.

dmbates:
eq. (10) in the current sources is different from that posted in the image and I think it is more nearly correct.  If does have $u^{(i)}$ in the left summand.  The only thing I would change in the current version is to replace u with delta in the linear combination that defines the right summand.  Currently the minimization is over delta but delta doesn't occur in the expression to be minimized.  But changing the u to a delta in that linear combination gets us back to the previous problem which is that the penalty is the squared length of u, not the squared length of delta.

I guess it is time for me to roll up my sleeves and go through the code to determine what it actually is that we are trying to explain.

dmbates:
I think there is a clue in this function defined in R/lmer.R
```
RglmerWrkIter <- function(pp, resp, uOnly=FALSE) {
    pp$updateXwts(resp$sqrtWrkWt())
    pp$updateDecomp()
    pp$updateRes(resp$wtWrkResp())
    if (uOnly) pp$solveU() else pp$solve()
    resp$updateMu(pp$linPred(1))        # full increment
    resp$resDev() + pp$sqrL(1)
}
```
This is an R version of another function in the C++ code, so this is sort-of the backstop.  Notice that it uses `resp$wtWrkResp()`.  In the C++ code for the `glmResp` class there are the definitions

```
    ArrayXd glmResp::wrkResids() const {
        return (d_y - d_mu).array() / muEta();
    }

    ArrayXd glmResp::wrkResp() const {
        return (d_eta - d_offset).array() + wrkResids();
    }
```

I think the situation is that we evaluate $u^{(i+1)}$ directly as the weighted least squares solution using the weighted working *response*, not the weighted working residual.  All this means is that we add the current `eta` (adjusted for an `offset` if there is one) back in to the residual.

That ensures the penalty is on the squared length of the spherical random effects, not of the increment to the spherical random effects.

@bbolker If you agree with this I will try to modify that paragraph in `glmer.Rnw` to describe what we actually do.

bbolker:
> @bbolker If you agree with this I will try to modify that paragraph in `glmer.Rnw` to describe what we actually do.

Yes please.

~~Note to self, the only other loose end is that I wanted to add a brief comment about choosing the number of quadrature points when doing AGQ (answer: try increasing numbers until the estimates stabilize, with warm-start if desired)~~ (this is done now)

dmbates:
It may be a sign that I am running out of steam but I chose to explain IRLS and PIRLS in a hand-waving sort of way.  I just said that the weights are inversely proportional to the variances of the responses at the current parameter values without going into detail on the derivative of the link function and the variance function associated with the family.
If those are thought to be necessary they could go into an appendix or in supplemental materials.  Unless you are planning to write your own implementation these details are usually not necessary.

Let me know if you think I have done to much excising of materials.

bbolker:
I hate to look a gift horse in the mouth, and we can go with this, but ... I'm a little disappointed not to be giving the details here. If we could figure out how equation 10 is really supposed to go I think (do you agree?) that everything else in the current presentation (before your last set of edits) was OK (with the possible exception of $i$-indexing, which I can sort out). I think of this paper as the companion to the 2015 JSS paper, where we actually give the gory details of the computation, so that there is a reference to what's actually going on other than the source code itself.

As I understand it, we know that eq. 13 is 'true' &mdash; it matches what we're actually doing in the software implementation. So one way of framing the question is, "what does eq. 10 have to look like in order for it to correspond to the condition given in (13)"?

Looking back at the corresponding discussion in *Embracing Uncertainty* ([here](https://embraceuncertaintybook.com/aGHQ.html#sec-PIRLS)), I see that it indeed a more shorthand description, closer to your new version.  

Is it true that the only missing link is the proper formulation of (10) to match (13) (which matches our code in `lme4` and in `lme4pureR`, and which almost certainly matches the implementation in `MixedModels.jl` as well ... ?

annahuynhly:
Hope I'm not interrupting; I made small changes. See comments below, corresponding to the changes:

Few comments:

1. Added UofT email since I'm active there, but technically not affiliated with UofT anymore (right now)... But for various reasons I think I'll be able to keep it active for a next couple of years

2. Z and b are not necessarily defined in the glmer paper. It is shown in section 2.3 but it was not given an introduction the same way as X and \beta was in section 2.1. (I edited this.)

This is a comment where I didn't implement anything myself, just wanted to clarify:
**3. Lines 92-93 (“Approximating the integral with respect to u of the joint density (evaluated at the observed y) by the integral of this local quadratic approximation is Laplace’s method.”) To clarify, one could also use AHQ as well, right? Unsure if we want to be specific and say that one could use Laplace/AHQ/any other method.**

4. I think it would be good to clarify that the argument nAGQ, the number of quadrature points, is simply just k in the equation between lines 219 and 220 in section 3... Wouldn’t it be better for there to be an equation number here as well? (I also edited this.)

5. Should we have a citation for Powell’s BOBYQA? (I edited this.)

6. Around lines 495-496, it is annoying that cs(1+urban|district, hom = TRUE) bleeds through the margins. (I edited this.)

bbolker:
@annahuynhly , all your edits seem sensible to me, I don't think we need to clarify about Laplace vs AGHQ at this point.

@dmbates, the other alternative if we feel like weaseling out of  correcting eq. 10 is to just avoid directly stating eq. 10, but give all the other detail: e.g. https://pages.stat.wisc.edu/~bates/PotsdamGLMM/GLMMD.pdf states the least-squares problem (eq 8 for us in the earlier version at 2eee9e896424) and the update (eq 13) without ever explicitly stating the definition of $\boldsymbol \delta$ directly ...

dmbates:
I agree that I was too terse in my hand-waving explanation of PIRLS.  My problem is that this is one of the calculations where the finicky details of numerical analysis are important.  It is not a calculation where you can hand an inexperienced research assistant (or a Chatbot) a set of formulas and say "go do it".  Especially for important cases like the Bernoulli the way that you evaluate things like the working weights can become very important for the edge cases.  The variance function for the Bernoulli is $\mu(1-\mu)$ where $\mu=1/(1+\exp(-\eta))$.  On the surface it appears that it should be possible to reverse the definition of "success" and "failure" and get the same answer, but not at the extremes.  You want to make sure that the variance doesn't round down to zero because the weights are inversely proportional to the variance.  For positive $\eta$ you round to zero at about $-\log(\epsilon)$ whereas for negative $\eta$ you have much more room to evaluate small $\eta$ without its rounding to zero.
```
julia> function varfunc(η)
           μ = inv(1. + exp(-η))
           return μ, μ * (1. - μ)
       end
varfunc (generic function with 1 method)

julia> log(eps())
-36.04365338911715

julia> varfunc(-37)
(8.533047625744066e-17, 8.533047625744065e-17)

julia> varfunc(37)
(1.0, 0.0)
```

The point I am trying to make is that implementation details matter and if we try to discuss all the implementation details it will make for a very long section indeed.

I think it is overly optimistic that we can write enough about the implementation that someone can implement it strictly from the description.  There were entire papers written about the initial implementation of `glm` in `R` and why it was slow and not terribly reliable.  `lme4` is open source so if someone really wants to find out about the implementation details they can read the sources.

Having said that, I agree that I was too terse and will revisit the references you mentioned.

dmbates:
Okay, I read the slides for the 2008 presentation by that Bates guy, who apparently doesn't know the term "canonical link".  What parts of the presentation do you think we should incorporate?  (I managed to confuse myself with the $\mathbf{T}(\mathbf{\theta})$ and $\mathbf{S}(\mathbf{\theta})$ matrices until I noticed that $\mathbf{T}(\mathbf{\theta})$ is defined to be **unit** lower triangular.)

We should formally define the penalized least squares problem, probably the version with $\mathbf{u}$ as the only variable being optimized.  I guess we can say that $\mathbf{P}$ is the fill-reducing permutation determined by the structure of $\mathbf{Z}$, then write out penalized least squares problem.  Can we just punt there and say that it is solved by taking the sparse Cholesky factor of the big expression on the left?

bbolker:
OK, one more try.  What do you think about [this attempt](https://github.com/lme4/lme4/blob/004049e0437b595d4aff2842073ef297bfcec0cc/misc/pirls.tex) (rendered PDF [here](https://github.com/lme4/lme4/blob/db/moresuggestedchanges/misc/pirls.pdf))?  This implements my suggestion above, i.e. including everything we had before except eq. 10, which as I understand it was the real sticking point. 

I take your point about implementation details, although in my opinion numerical accuracy/logspace computation issues are more or less orthogonal to the stuff we're struggling with here ...

bbolker:
One more question about the development here (sorry!)  It looks like the description of the algorithm (steps 1-3), $\boldsymbol W$ represents the full weight matrix (combination of prior and working weights), while in eqs 3 and 5 (as rendered in this mini-section) it is represents the prior weights only, i.e. $\boldsymbol W^{\textrm{old}} = \boldsymbol M^{(i)} \boldsymbol W \boldsymbol M^{(i)}$ ... ? (Even though $\boldsymbol M^\top = \boldsymbol M$, because $\boldsymbol M$ is diagonal, is there any point to writing this as $\boldsymbol M^{(i)\top} \boldsymbol W \boldsymbol M^{(i)}$ where it occurs, so it's a little less jarring? (I don't know which way causes more/less cognitive dissonance ...)

dmbates:
Do you know that the code actually implements what is described in the pirls.pdf document?  My reading of the code is that it is using a weighted, working response in the penalized reweighted least squares problem, similar to what I described in Appendix C of https://EmbraceUncertaintyBook.com, especially section C.4

I should have been more explicit about that.  I think we want to write the solution like equation C.8 and the following unnumbered equation in Embrace Uncertainty.  That is, we use a sparse Cholesky factor to solve
$\left(\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\mathbf{Z}\mathbf{\Lambda} + \mathbf{I}\right)\mathbf{u}^{(k+1)}=\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\tilde{\mathbf{y}}^{(k)}$
where $\tilde{\mathbf{y}}^{(k)}$ is the working response at iteration $k$.

I don't think that what is shown as equation (3) in the pirls.pdf document is what is being done in the code, which is why I cut that part out.

I am sorry that I have been slow about responding on some of these changes.  It is taking me a while to reacquaint myself with the lme4 code.

bbolker:
I will look again tomorrow ... what is the definition of $\mathbf{\tilde y}$ in the *Embracing Uncertainty* appendix?

dmbates:
It is just before section C.3.1 .  Take the working residual and add back in $\mathbf{\eta} - \mathbf{o}$ where $\mathbf{o}$ is the offset.  The offset part is important in PIRLS because that is where the $\mathbf{X}\mathbf{\beta}$ contribution is incorporated when optimizing with respect to $\mathbf{u}$.

dmbates:
I am just working out the LaTeX glitches in a detailed explanation of PIRLS.  I hope to push a revision within an hour.

dmbates:
I have pushed the changes adding a more detailed explanation of PIRLS.  I will now go back to the code to ensure that what I wrote reflects what is actually done.  When reading this explanation watch for my getting things backwards like which side the permutation matrix occurs on and which side has its transpose.

dmbates:
In the file `src/respModule.cpp` around lines 125-160 several extractors and modifiers are defined for a `glmResp` struct.  Because a `glmResp` includes an `lmResp` updates that affect the `lmResp` parts only, like `updateMu` (which just adds on the offset to the $\mathbf{X\beta}$ part) are defined in the earlier section.

The real action takes place in the R-callable `pwrssUpdate` defined around line 308 of `src/external.cpp`, which in turn relies on `internal_glmerWrkIter` from around line 268 in that file.  That function updates the diagonal weights matrix $\mathbf W$ and the decomposition $\mathbf L$ then solves for $\mathbf u^{(i+1)}$ from the weighted working response (lines 278-280).

I feel confident that what I wrote is pretty close to what actually happens, after adjusting for the typos and thinkos about where the transposes go.

I believe that the R function `mkdevfun` in `R/lmer.R` sets up this evaluation but I could be wrong.  I don't claim authorship of `mkdevfun`.

bbolker:
I'm **extremely** ready to be wrong, but alas I'm still not convinced. I can see that the action is `internal_glmerWrkIter`, and I can see that various stuff is being updated; my challenge at the moment is to figure out *what linear system is actually being solved*.

To be clear, we have reduced this to a question of which of two formulations is actually what's done in the code:

* (Bates, modern)

$$\left(\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\mathbf{Z}\mathbf{\Lambda} + \mathbf{I}\right)\mathbf{u}^{(k+1)}=\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\tilde{\mathbf{y}}^{(k)}$$

where $\tilde{\mathbf{y}}^{(k)}$ is the working response at iteration $k$.

* **or** (what was written in 2015)

$$\left(\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\mathbf{Z}\mathbf{\Lambda} + \mathbf{I}\right)\mathbf{\delta}^{(k)}=\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W} (\mathbf y - \mathbf \mu^{(k)}) - \mathbf{u}^{(k)}$$

(in both cases I'm assuming we're absorbing factors of $d\mathbf {\mu}/d\mathbf{\eta}$, or $\mathbf J$ in DB's notation, into $\mathbf W$). So the differences are in (1) what we're solving *for* (is it the value of $\mathbf u$ at the next step, or just the increment of $\mathbf u$ (i.e. $\mathbf \delta$) from one step to the next?) and (2) whether we have $\tilde{\mathbf{y}}^{(k)}$ or $(\mathbf y - \mathbf \mu^{(k)}) - \mathbf{u}^{(k)}$ on the RHS (I'm being a tiny bit sloppy here, the $(\mathbf y - \mathbf \mu^{(k)})$ is multiplied by $\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}$ but the $\mathbf u^{(k)}$ term isn't)

Here's the meat of `internal_glmerWrkiter`, without comments:
```c++
pp->updateXwts(rp->sqrtWrkWt());
pp->updateDecomp();
pp->updateRes(rp->wtWrkResp());
if (uOnly) pp->solveU();
else pp->solve();
```

So if we really want to know what system is being solved, we have to look at the `solveU` method (which is what we use if nAGQ>0):

```c++
merPredD::Scalar merPredD::solveU() {
  d_delb.setZero(); // in calculation of linPred delb should be zero after solveU
  // BMB: d_Utr is d_LamtUt * wtres (predModule.cpp:307)
  // BMB: wtres is sqrtrwt * (y-mu) (respModule.cpp:51ff; this is the working residual)
  d_delu          = d_Utr - d_u0;   
  d_L.solveInPlace(d_delu, CHOLMOD_P);
  d_L.solveInPlace(d_delu, CHOLMOD_L);    // d_delu now contains cu
  d_CcNumer       = d_delu.squaredNorm(); // numerator of convergence criterion
  d_L.solveInPlace(d_delu, CHOLMOD_Lt);
  d_L.solveInPlace(d_delu, CHOLMOD_Pt);
  return d_CcNumer;
}
```
I am having a hard time digging far enough down to see what's inside `dL`, but I'm guessing it's equivalent to $\left(\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\mathbf{Z}\mathbf{\Lambda} + \mathbf{I}\right)$ (which is the part that we agree on anyway)

`d_delu` above looks a lot like what we have for the RHS in version 2 (`d_LamtUt * sqrtwt * (y-mu) - d_u0`). I can't tell what we are doing with the quantity we're solving for (i.e. whether we're really solving for $\mathbf \delta^{(k)}$ or $\mathbf u^{(k+1)}$, but the fact that it's called `delu` throughout definitely makes me think it's an increment rather than a next-step value ...

The other line of reasoning is looking at `lme4pureR`. I stipulate that Steve Walker might have written the code wrong in the first place &mdash; that is, if he was working from the embryonic glmer JSS paper rather than from the C++ source code, he would naturally have set things up in that way &mdash; but I can confirm that the *results* from fitting the same model with `glmer` and `lme4pureR` (see [here](https://github.com/lme4/lme4pureR/blob/master/tests/pirls.R#L1-L25)) match more closely (parameters within 5e-6, log-likelihood within 5e-8) than I would expect if they were setting up the computation differently ... and the body of the `lme4pureR` PIRLS definitely matches version 2 above ...

```r
Whalf <- Diagonal(x = sqrt(weights / variance(mu)))
LtZtMWhalf <- LtZt %*% (Diagonal(x = muEta(eta)) %*% Whalf)
L <- update(L, LtZtMWhalf, 1)
## alternative (more explicit but slower) Cholesky update
# L <- Cholesky(tcrossprod(LtZtMWhalf), perm=FALSE, LDL=FALSE, Imult=1)
wtres <- Whalf %*% (y - mu)
# solve for the increment
delu <- as.vector(solve(L, LtZtMWhalf %*% wtres - u))
```

I'd be happy to have a (video) synchronous/face-to-face chat about this &mdash; seems like it might be the most efficient way forward at this point ... ???

(It has also crossed my mind that these two formulations might in fact be exactly equivalent if we thought about it carefully enough, but I haven't been able to make them match yet ...)

PPP...S  We should check whether it's supposed to be a factor of $\mathbf W$ or $\mathbf W^{1/2}$ on the RHS of the equation (either one) ...

I also took a quick look at `pirls!` in `MixedModels.jl` (not that it would necessarily be using the same formulation), but couldn't decipher it quickly ...

dmbates:
To me the definitive line in the code of `internal_glmerWrkiter` is updating the "residual", which in this case means the right-hand side - the "residual" name is vestigial, with `rp->wtWrkResp()`.  It is setting up to solve a penalized weighted least squares problem.  If you have a working response on the right hand side or a weighted working response, $\mathbf W^{1/2}\tilde{\mathbf{y}}$, then you are solving for the next value of $\mathbf u$.

The fact that this differs from the flow of the code in `lme4pureR` may have to do with when `lme4pureR` was written.  I have looked at the output of `git blame` on some of the C++ source and the major changes were made about 15 years ago by me.  I think we were still using svn at the time but the commits and the commit comments are recorded in the git archive.  The section of `src/external.cpp` near the definition of the `wtWrkResp` method was modified by @mmaechler about 14 years ago but his comment says that it was a fix from Doug.

I am willing to go through a detailed examination of the values produced in one or more examples so we can check at every point what the values being produced are and what we think they should be based on other code that we understand better. I would prefer to do such a cross-checking of code with my writing in Julia.   I really don't want to go back to the C++ code if I can avoid it.  To do debugging on the C++ code would require relearning C++, Rcpp, gdb (or whatever other symbolic debugger is in fashion these days), R interfacing, etc. and that is not really how I want to spend my retirement.

That section of the paper from 2015 looks like I had a hand in writing it.  I have a vague recollection of talking with Peter McCullaugh around then and changing from evaluating an increment, based on a right-hand side which is a working residual, to evaluating an updated parameter value, based on a right hand side that is a working response.  He said that these were equivalent.  The reason it is important to evaluate an updated parameter value is that makes it possible to incorporate the penalty by inflating the diagonal of the matrix going in to the sparse Cholesky.  I'm not sure if we use it but the CHOLMOD code allows you to do that with a simple scalar argument to the function that does the numerical factorization.

Does it seem as if it would be possible to take the different examples and tear apart the stages of the update until we can reproduce them in other, easier to understand, code?

dmbates:
If you check https://github.com/lme4/lme4/commit/01e87f6ee149768a622e72222ef91a11572a86e9 which was svn commit by Martin in April of 2012 around line 350 of `src/external.cpp` this is when we switched from the weighted working residual to the weighted working response in the `glmerWrkIter` externally-callable function.

dmbates:
Anything further I can help with?

bbolker:
Thanks. I'm working on some stuff now, will post shortly.

bbolker:
I've added a [tiny example](https://github.com/lme4/lme4/blob/db/moresuggestedchanges/misc/explore_pirls.R) (9 observations total, 3 groups, so the latent vector is only three elements long), and (with Claude's help) fixed a few bugs so that we can debug mostly in R by setting `control=glmerControl(compDev=FALSE)`. (`compDev=FALSE` does more of the work in R, calling `RglmerWrkIter`, and gives identical results to the full-C++ call.)

Despite the example being very small, it nevertheless gets through the fitting procedure without any obvious problems.

 The algorithm still falls through to to C++ at the `pp$solveU()` step, but we can access any of the fields in the `pp` object (a `merPredD` object) from R, including `pp$Utr`, `pp$u0`, `pp$Xwts`, and `pp$delu`, interactively ...

It should now be reasonably straightforward to step through a single iteration of the PIRLS algorithm to see what's happening ... ???

(I was initially going to suggest working on this by enabling existing debug-print statements in the C++ code, but working via `compDev = FALSE` should be much more convenient ...)

bbolker:
Okay, I may have some clarity/a resolution now. Bottom line:

* we are indeed solving for $\mathbf u^{(i+1)}$, not $\boldsymbol \delta^{(i)}$
* With the exception of a small question about weighted vs. non-weighted working responses (see first bullet point below) I agree that your current version of the ms. is correctly framed. 

The gory details of stepping through a single PIRLS iteration (starting from $\mathbf u = (0,0,0)$, which does have the potential to be non-generic ...) are [here](https://github.com/lme4/lme4/blob/db/moresuggestedchanges/misc/explore_pirls.R), if you want to see.

Here are my further thoughts about the state of things (current version of the ms. and my understanding):

* I'm still trying to figure out if we need a factor of $\mathbf W^{1/2}$ on the RHS of eq. 15 (which is currently written as 

$$
{\boldsymbol\Lambda}^\top{\mathbf Z}^\top{\tilde{\mathbf y}^{(0)}}
$$

). *Embracing Uncertainty* has 

$$
{\boldsymbol \Lambda}^\top {\mathbf Z}^\top {\mathbf W} \tilde{\mathbf y}
$$ 

on the RHS; my guess is that should actually be  $\mathbf W^{1/2}$ ... the code has `LamtUt %*% resp$wtWrkResp())`, and `$wtWrkResp()` is (not surprisingly) the *weighted* working response ...
* There are still some aspects of the internal 'notation' that puzzle me, e.g. I'm not quite sure what `u0` is doing, but I'm ready to let it go ... (in a perfect world we would work out the correspondence between equations in the ms. and code in the package, and comment the code appropriately &mdash; and have  a table like Table 7 of the JSS paper). It **might** be worth adding a comment to the paper saying that the efficient method used internally for solving the equation is similar to that used for solving the normal equations in the LMM case, as described in equations 49-53 of the JSS paper (this helped me to understand `pp$solveU()` a little bit better ...)
* in the C++ machinery (specifically the `updateDecomp` method of `merPredD` objects), we're updating `L`, `RX` and `RZX`, which aren't explicitly mentioned anywhere in our text. Presumably these are defined as in eq. 18 of the lmer JSS paper, i.e.  `L` = upper left block of (lower-tri) Cholesky factor; `RXt` = lower right block; `RZXt` = lower left block ... (is there a parallel to this structure in the Julia code, or did you find a different way to set things up?)

I think that's it. We're sooooo close now ...

bbolker:
PS @dmbates : since we no longer have a 'minimizing equation' to refer back to, is this a sufficiently accurate hand-wavy description of what we do with `nAGQ=0` (i.e., is the parenthetical statement OK)?

> profiles out the fixed effects by minimizing with respect to $\boldsymbol\beta$ and the conditional modes $\mathbf u$ simultaneously in the PIRLS algorithm (i.e., replacing $\mathbf u$ with $(\boldsymbol \beta \mathbf u)$ and $\mathbf Z \boldsymbol \Lambda$ with $(\mathbf X | \mathbf Z \boldsymbol \Lambda)$ throughout, except with the penalty term still applying only to $\mathbf u$)

dmbates:
> PS @dmbates : since we no longer have a 'minimizing equation' to refer back to, is this a sufficiently accurate hand-wavy description of what we do with `nAGQ=0` (i.e., is the parenthetical statement OK)?
> 
> > profiles out the fixed effects by minimizing with respect to   β and the conditional modes    u simultaneously in the PIRLS algorithm (i.e., replacing  u with (β  u) and  ZΛ with (X  |    Z Λ ) throughout, except with the penalty term still applying only to  u )

I think I would just say "minimizing with respect to β and u simultaneously".  To me the "conditional modes" are the minimizing values, not the variable over which the minimization is being done.  Other than that, this looks fine.  Thanks for catching that.

bbolker:
@dmbates in case you missed it, see comment/question above about $\mathbf W$ vs $\mathbf W^{1/2}$, among all the other blather &mdash; it's the only bit I really need an answer to ...

dmbates:
> @dmbates in case you missed it, see comment/question above about W vs W 1 / 2 , among all the other blather — it's the only bit I really need an answer to ...

Is $\mathbf{U}^\top = \mathbf{Z}^\top\mathbf{W}^{1/2}$?  The right hand side should be 

$$
\mathbf{\Lambda}^\top\mathbf{Z}^\top\mathbf{W}\tilde{\mathbf{y}}
$$

I think for convenience I wrote the extractor in the C++ code to return the weighted working response, 

$$
\mathbf{W}^{1/2}\tilde{\mathbf{y}}
$$

so there should be a $\mathbf{W}^{1/2}$ post-multiplying $\mathbf{Z}^\top$.

dmbates:
@bbolker When I run the first part of `lme4/misc/explore_pirls.R` I get slightly different values of `VarCorr(g1)` (1.162) and `VarCorr(g2)` (1.1596).  Do I need to re-install lme4?

bbolker:
Yes, you do need to re-install `lme4` (from the `db/moresuggestedchanges` branch) - it was [this fix](https://github.com/lme4/lme4/commit/c174377fce06dd3e061f21904bbcb276694541d0), I believe.

And yes, you're right that there's a $\mathbf W^{1/2}$ in $\mathbf U$ and another one needed for $\tilde {\mathbf y}$, so the version in *Embracing Uncertainty* (with $\mathbf W$) is correct; I'll edit eq. 15 accordingly.

PS what do you think about

> adding a comment to the paper saying that the efficient method used internally for solving the equation is similar to that used for solving the normal equations in the LMM case, as described in equations 49-53 of the JSS paper

...?

bbolker:
From Rune:

> 1. In line 599 [now 604], I think beta should be u if I understand you correctly

edited.

> 2. Eq 27: The last u^\top does not follow from the elements above - perhaps a term is missing from the Eq. just below line 594?

The $u^\top$ comes from the $-\frac{1}{2} \mathbf u^\top \mathbf u$ term in eq. 25. I think those were dropped from the equations in 7.1 because we're treating $L$ as the standard (unpenalized) GLM log-density, right? How do we denote this? Should we use a different symbol ($L'$?) to denote the part of the density not including the penalty term, or include the additional term throughout? What's the least clunky way to do this?

Should line 602 read "The only difference induced by subtracting the  $\frac{1}{2} \mathbf u^\top \mathbf u$ term ... ?" (or am I misunderstanding?)

> 3. In Eq. 18 I am not sure what \tilde f is - perhaps put a name on it in line 176?

I added "(where $\tilde f$ is the penalized deviance evaluated at the conditional modes)" [now line 180]. Is that OK?

> 4. It is my understanding that the second order derivative in Eq. 32 is related to the L-matrix in Eq.s 18 and 19. We also have the approximation in Eq 33. For canonical link functions Eq.s 32 and 33 are the same. I think it would be worth mentioning if for non-canonical link functions, the approximate Laplace method using Eq. 33 is used or if the full Laplace method (Eq. 32) is used.

?? Not sure ?? Are we (like `glm`, IIUC) using Fisher scoring throughout, because the Fisher approx takes us to the same MLE as the eq 32 ?

dmbates:
> From Rune:
> 
> > 1. In line 599 [now 604], I think beta should be u if I understand you correctly
> 
> edited.


I don't see a beta in line 599 or line 604 (I installed the latest `lme4/master` and re-ran `knitr::knit2pdf("glmer.Rnw")`
to check.  The only `\beta` that I see is in expressions like `dL(\beta,\theta|\mathbf{y},\mathbf{u})`.  Does this mean that you have already fixed it?

> > 2. Eq 27: The last u^\top does not follow from the elements above - perhaps a term is missing from the Eq. just below line 594?
> 
> The u ⊤ comes from the − 1 2 u ⊤ u term in eq. 25. I think those were dropped from the equations in 7.1 because we're treating L as the standard (unpenalized) GLM log-density, right? How do we denote this? Should we use a different symbol ( L ′ ?) to denote the part of the density not including the penalty term, or include the additional term throughout? What's the least clunky way to do this?

I agree that the $\mathbf{u}^\top$ comes from the $-\frac{1}{2}\mathbf{u}^\top\mathbf{u}$ in eq. 25.  I think we will need to add $-\frac{1}{2}\frac{d(\mathbf{u}^\top\mathbf{u}}{d\mathbf{u}}$ in the equation following line 602.

> Should line 602 read "The only difference induced by subtracting the 1 2 u ⊤ u term ... ?" (or am I misunderstanding?)
> 
> > 3. In Eq. 18 I am not sure what \tilde f is - perhaps put a name on it in line 176?
> 
> I added "(where f ~ is the penalized deviance evaluated at the conditional modes)" [now line 180]. Is that OK?

Seems good to me.
> 
> > 4. It is my understanding that the second order derivative in Eq. 32 is related to the L-matrix in Eq.s 18 and 19. We also have the approximation in Eq 33. For canonical link functions Eq.s 32 and 33 are the same. I think it would be worth mentioning if for non-canonical link functions, the approximate Laplace method using Eq. 33 is used or if the full Laplace method (Eq. 32) is used.

I really don't know.  To tell the truth I didn't go over the material in the Appendix carefully.  I don't think I had a hand in writing that and I find approaching GLMs through a very general formulation of the exponential dispersion family to be more confusing than illuminating.  Do we really need the appendix?

> ?? Not sure ?? Are we (like `glm`, IIUC) using Fisher scoring throughout, because the Fisher approx takes us to the same MLE as the eq 32 ?

bbolker:
> > From Rune:
> > > 1. In line 599 [now 604], I think beta should be u if I understand you correctly
> > 
> > 
> > edited.
> 
> I don't see a beta in line 599 or line 604 (I installed the latest `lme4/master` and re-ran `knitr::knit2pdf("glmer.Rnw")` to check. The only `\beta` that I see is in expressions like `dL(\beta,\theta|\mathbf{y},\mathbf{u})`. Does this mean that you have already fixed it?

  Yes, already fixed.

> 
> > > 2. Eq 27: The last u^\top does not follow from the elements above - perhaps a term is missing from the Eq. just below line 594?
> > 
> > 
> > The u ⊤ comes from the − 1 2 u ⊤ u term in eq. 25. I think those were dropped from the equations in 7.1 because we're treating L as the standard (unpenalized) GLM log-density, right? How do we denote this? Should we use a different symbol ( L ′ ?) to denote the part of the density not including the penalty term, or include the additional term throughout? What's the least clunky way to do this?
> 
> I agree that the u ⊤ comes from the − 1 2 u ⊤ u in eq. 25. I think we will need to add − 1 2 d ( u ⊤ u d u in the equation following line 602.
> 
> > Should line 602 read "The only difference induced by subtracting the 1 2 u ⊤ u term ... ?" (or am I misunderstanding?)

I've added some stuff here, but now I'm worried about a sign inconsistency. If (25) is really the log-density (and not the negative log-density), then the penalty term is correctly negative ($-\frac{1}{2} \mathbf u \mathbf u^\top$). However, does that mean that the last term in (27), and all of the $\mathbf I_q$ terms in the subsequent Hessian expressions, should also be negative ... ??

The sign confusion looks pervasive. We are working with the conditional log-density (not the negative log-density) throughout, so if we use $-\mathbf I_q$ throughout, then we're good until we get to equations 33 and 34, where the sign of the first ($\mathbf U^\top \mathbf U$) term magically flips from negative to positive between the second and third expression.

I have (I think) fixed everything so it's correct and consistent, but please double-check ...

> > > 3. In Eq. 18 I am not sure what \tilde f is - perhaps put a name on it in line 176?
> > 
> > 
> > I added "(where f ~ is the penalized deviance evaluated at the conditional modes)" [now line 180]. Is that OK?
> 
> Seems good to me.
> 
> > > 4. It is my understanding that the second order derivative in Eq. 32 is related to the L-matrix in Eq.s 18 and 19. We also have the approximation in Eq 33. For canonical link functions Eq.s 32 and 33 are the same. I think it would be worth mentioning if for non-canonical link functions, the approximate Laplace method using Eq. 33 is used or if the full Laplace method (Eq. 32) is used.
> 
> I really don't know. To tell the truth I didn't go over the material in the Appendix carefully. I don't think I had a hand in writing that and I find approaching GLMs through a very general formulation of the exponential dispersion family to be more confusing than illuminating. Do we really need the appendix?

  I'd like to keep it if we can.

> 
> > ?? Not sure ?? Are we (like `glm`, IIUC) using Fisher scoring throughout, because the Fisher approx takes us to the same MLE as the eq 32 ?

I think we must be using Fisher scoring if we're using PIRLS. Added a statement to that effect.

dmbates:
I can follow through the gradient and I think that part is fine.  In the Hessian part I have trouble adjusting to the signs because I implicitly think of minimization problems, not maximization, and I want the Hessian to be positive definite, not negative definite.  Once we get to the Hessian I can't vouch for the formulas as I have trouble keeping track of all the pieces.

I think of IRLS and PIRLS as using a local linear approximation to the link to give a weighted least squares problem to solve. For me Fisher scoring doesn't enter into the picture.

I wish I could be more helpful but I have reached the "old dog, new tricks" stage of life.

bbolker:
That's fair. If it doesn't seem clearly **wrong** to you then I think we'll call it Done (the old "well the reviewers will let us know if it's wrong" cop-out ...) I'll also ask @annahuynhly to take one last look at the appendix before submitting.

Here we go!

annahuynhly:
I'll try to look at it tonight, or worst case during the weekend. My new job is super demanding, but I'm committed.

